PR #20, opened 2018-03-20
adding runtime configuration structure

### LINKED ISSUE #5: Make Runtime interface accept interfaces
Make Runtime interface accept interfaces to allow users to completely override all of a runtime's behavior.

- refactor interface to remove hook calls
- create config type struct for extracted credentials
- create config type struct for sandbox hooks

shawn-hurley:
* adding pre sandbox hooks
fixes #5


jmrodri:
Test coverage increased. Runtime increased from 10.5% to 21.0%

```diff
--- b	2018-03-20 20:56:43.874972243 -0400
+++ o	2018-03-20 20:55:09.185916365 -0400
@@ -1,6 +1,7 @@
-ok  	github.com/automationbroker/bundle-lib/apb	0.063s	coverage: 4.1% of statements
-ok  	github.com/automationbroker/bundle-lib/clients	0.044s	coverage: 20.8% of statements
+ok  	github.com/automationbroker/bundle-lib/apb	0.049s	coverage: 4.1% of statements
+ok  	github.com/automationbroker/bundle-lib/clients	0.040s	coverage: 20.8% of statements
 ?   	github.com/automationbroker/bundle-lib/crd	[no test files]
-ok  	github.com/automationbroker/bundle-lib/registries	0.034s	coverage: 74.7% of statements
-ok  	github.com/automationbroker/bundle-lib/registries/adapters	0.082s	coverage: 6.0% of statements
-ok  	github.com/automationbroker/bundle-lib/runtime	0.038s	coverage: 10.5% of statements
+ok  	github.com/automationbroker/bundle-lib/registries	0.048s	coverage: 74.7% of statements
+ok  	github.com/automationbroker/bundle-lib/registries/adapters	0.090s	coverage: 6.0% of statements
+ok  	github.com/automationbroker/bundle-lib/runtime	0.039s	coverage: 21.0% of statements
+?   	github.com/automationbroker/bundle-lib/runtime/mocks	[no test files]
```

jmrodri (approved):
ACK

runtime/runtime.go
```diff
@@ -50,7 +57,9 @@ type provider struct {
 	coe
 	ExtractedCredential
 	postSandboxCreate  []PostSandboxCreate
+	preSandboxCreate   []PreSandboxCreate
```
rthallisey, line 64: We have multiple functions provided as types. I'd rather use an interface ```SandboxHooks``` and encapsulate all of them.

shawn-hurley: How exactly would that work? Could you give a code snippet because I am not following what you mean?

rthallisey: Exactly like ```ExtractedCredential```.
```go
type SandBoxHooks interface {
     PostSandboxCreate(...)
     PreSandboxCreate(...)
     PostSandboxDestroy(...)
     PreSandboxDestroy(...)
}
```

shawn-hurley: But that is not how this works right, we want to add an arbitrary number of hooks? I don't understand how this interface works in the context of how we are using the hooks.

rthallisey:
```go
type SandBoxHooks interface {
     postSandboxCreate(...)
     preSandboxCreate(...)
     postSandboxDestroy(...)
     preSandboxDestroy(...)
}

type Configuration struct {
        SandBoxHooks
	ExtractedCredential
}

Func NewRuntime(config Configuration) {

        ...

        var c ExtractedCredential
        if config.ExtractedCredential == nil 
                c = defaultExtractedCredential{}
        } else {
                c = config.ExtractedCredential
        }

        var h SandBoxHooks
        if config.SandBoxHooks == nil 
                h = cluster.SandBoxHooks
        } else {
                h = config.SandBoxHooks
        }

        Provider = &provider{coe: cluster, ExtractedCredential: c, Hooks: h}
}

func (p provider) CreateSandbox(...) {

...
         p.Hooks.postSandboxCreate(...)
...
}
```

rthallisey: The crux of this argument comes down to how users primarily use this API.  Will they go with what a provider has then append to it or do they want to overwrite everything. Right now, we only have one hook and it's in the openshift provider so appending is fine.  To me, this solution is functional, but I think we could come back to this if we hit a limitation with append, but it's too soon to tell if we will.  

My favorite option is still to use the interface and maybe create a core set of hooks enabled by flags, but I'll go with what folks think on this.  I don't want to block this work from proceeding.

rthallisey (approved):

Issue #6, opened 2018-03-15
Rename APB to Bundle


djzager:
Only make public what needs to be made public, the rest should be private.

shawn-hurley:
Need to make ExtractCredentials, ExtractCredentialsAsFile, ExtractCredentialsAsSecret unexported and then I think this can be closed

shawn-hurley:
because of #7 we are just going to move those functions to runtime. This should be closed then with #32

PR #32, opened 2018-03-21
initial rename of apb to bundle

shawn-hurley:


shawn-hurley:
partial fix for #6 will allow us to change the names in the broker now

jmrodri:
I would do this AFTER the broker has vendored bundle-lib.  I think it will be easier on both sides.

dymurray:
@shawn-hurley these changes look good to me, only concern is I noticed all the tests are now testing images named `foo-bundle`. Is the intention that the whitelist/repo names all change too? I know that is outside the scope of this PR just want your thoughts.

dymurray (approved):

shawn-hurley:
@dymurray my thoughts on this are:
1. It makes sense that APB -> bundle in the library to me. I think the library should be as specific type of bundle agnostic as possible.
2. I think the automation broker should stay the exact same. it assumes APBs right now. I think that if an admin wants to name their stuff bundle and use bundle go for it. 
3. I think that we already have a lot of APBs out there and I don't see a reason to change those, and I was not attempting to advocate for that with this change.

shawn-hurley:
@jmrodri has approved the PR, talked to him over chat and I going to merge now.

PR #40, opened 2018-03-23
Add helm registry adapter

djzager:


Gopkg.lock
```diff
@@ -215,6 +230,22 @@
...
   version = "1.0.0"
 
+[[projects]]
+  name = "github.com/op/go-logging"
+  packages = ["."]
+  revision = "b2cb9fa56473e98db8caba80237377e83fe44db5"
+  version = "v1"
+
+[[projects]]
+  branch = "bind-wkflw"
```
shawn-hurley: Please remove this.

Gopkg.lock
```diff
@@ -215,6 +230,22 @@
   revision = "1df9eeb2bb81f327b96228865c5687bc2194af3f"
   version = "1.0.0"
 
+[[projects]]
+  name = "github.com/op/go-logging"
```
shawn-hurley: Please remove this

Gopkg.lock
```diff
@@ -57,6 +63,15 @@
   revision = "346938d642f2ec3594ed81d874461961cd0faa76"
   version = "v1.1.0"
 
+[[projects]]
+  name = "github.com/docker/distribution"
```
shawn-hurley: It would be nice if we didn't have to vendor this

Gopkg.lock
```diff
@@ -546,13 +581,85 @@
...
+    "pkg/client/clientset_generated/clientset/typed/rbac/v1alpha1",
+    "pkg/client/clientset_generated/clientset/typed/rbac/v1beta1",
+    "pkg/client/clientset_generated/clientset/typed/settings/v1alpha1",
+    "pkg/client/clientset_generated/clientset/typed/storage/v1",
+    "pkg/client/clientset_generated/clientset/typed/storage/v1beta1",
+    "pkg/util",
+    "pkg/util/parsers"
+  ]
+  revision = "8c7c1c8b0ce0866d5ded2ce4fa402a716ce6bb6c"
+  version = "v1.7.15"
```
shawn-hurley: This needs to not be this version. we should be using 1.9.X

shawn-hurley (changes requested):
The build is failing because of the below comments

djzager:
Thank you for the helm @shawn-hurley I discovered that I had just copy/pasted the broker version of the helm-adapter into this one...forgetting to update my imports in the process. That was my problem, fixed the imports then ran `make ensure` and `make build` :sunglasses:

djzager:
helps if I update the tests too.

registries/adapters/helm_adapter.go
```diff
@@ -0,0 +1,326 @@
...
+		charts, ok := r.Charts[name]
+		if !ok {
+			continue
+		}
+
+		for _, chart := range charts {
+			chartVersions = append(chartVersions, chart.Version)
+		}
+		// Use the latest chart for creating the bundle
+		chart := charts[0]
```
shawn-hurley: could charts be empty?

shawn-hurley (approved):
Might need to address the comment above

registries/adapters/helm_adapter.go
djzager: I can add an additional check, but we currently protect against this by throwing away charts from the index that don't have at least one chart.

https://github.com/djzager/bundle-lib/blob/4634b97c2d8aed4ccc1a8573f97aa56b98fcd24e/registries/adapters/helm_adapter.go#L154-L157

dymurray (approved):

registries/adapters/helm_adapter.go
```diff
@@ -0,0 +1,326 @@
...
+type HelmAdapter struct {
+	Config Configuration
+	Charts map[string]ChartVersions
+}
+
+// ChartVersions is a list of versioned chart references.
+// Implements a sorter on Version.
+type ChartVersions []*ChartVersion
+
+// Len returns the length.
```
jmrodri: length of what? number of chart versions? Since ChartVersions is a type, we don't really know it's a slice underneath.

registries/adapters/helm_adapter.go
```diff
@@ -0,0 +1,326 @@
...
+func (c ChartVersions) Len() int { return len(c) }
+
+// Swap swaps the position of two items in the versions slice.
+func (c ChartVersions) Swap(i, j int) { c[i], c[j] = c[j], c[i] }
```
jmrodri: Hey that's simpler than adding and subtracting them :)

registries/adapters/helm_adapter.go
jmrodri: I would add a comment that explicitly states that we're okay because we've thrown away charts that didn't have at least one chart. Then we don't come back a year later and wonder hrm does this work? :)

registries/adapters/helm_adapter.go
```diff
@@ -0,0 +1,326 @@
...
+			},
+		}
+
+		specs = append(specs, spec)
+	}
+
+	return specs, nil
+}
+
+// getHelmIndex returns an helm repository IndexFile object
```
jmrodri: typo: an helm repository to a helm repository.

I realize there is some debate on whether a vs an in front of h depending on whether the h is silent (as in hour) or pronounced as in hello. I think 'a helm repository' sounds more correct.

registries/adapters/helm_adapter.go
```diff
@@ -0,0 +1,326 @@
...
+		sort.Sort(sort.Reverse(versions))
+	}
+	if index.APIVersion == "" {
+		return index, fmt.Errorf("No APIVersion on Index file")
+	}
+
+	return index, nil
+}
+
+// loadArchive returns Helm Chart values as a string
```
jmrodri: typo: returns Helm Chart values -> returns the Helm Chart values

jmrodri (approved):
There are some typos in the comments which are not critical, but should get fixed at some point.

registries/adapters/helm_adapter.go
djzager: I like that you defended your position.

registries/adapters/helm_adapter.go
eriknelson: Surprised this passes the linter , but I like it :+1:

registries/adapters/helm_adapter.go
```diff
@@ -0,0 +1,326 @@
...
+type ChartVersions []*ChartVersion
+
+// Len returns the length.
+func (c ChartVersions) Len() int { return len(c) }
+
+// Swap swaps the position of two items in the versions slice.
+func (c ChartVersions) Swap(i, j int) { c[i], c[j] = c[j], c[i] }
+
+// Less returns true if the version of entry a is less than the version of entry b.
+func (c ChartVersions) Less(a, b int) bool {
```
eriknelson, line 59: :+1:

jmrodri:
ACK

eriknelson (approved):

PR #42, opened 2018-03-26
Update OpenShift registry to use www-auth header

dymurray:
* hit `<registry_route>/v2/` and check the returned `www-authenticate` header to get bearer token URL
* Contact URL from step one with basic auth to retrieve bearer token
* Use in requests to get manifests

### DIFF
--- a/registries/adapters/openshift_adapter.go
+++ b/registries/adapters/openshift_adapter.go
@@ -23,8 +23,8 @@ import (
 	"net/http"
 	"strings"
 
-	logging "github.com/op/go-logging"
-	"github.com/openshift/ansible-service-broker/pkg/apb"
+	"github.com/automationbroker/bundle-lib/apb"
+	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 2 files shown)


shawn-hurley (approved):
LGTM w/ the adding the test case

djzager (changes requested):
I'm with @shawn-hurley I think a unit test for this would be great. I think the `rhcc` adapter has a good unit test, if you need another example, there is one in [my pr](https://github.com/automationbroker/bundle-lib/pull/40/files#diff-dcc0d02d4bf32f405d79f21da9da733c).

eriknelson (changes requested):
Waiting on test case.

dymurray:
@eriknelson @djzager working on test cases.

djzager (approved):
Thanks for going through that pain. LGTM

shawn-hurley:
Merging because we were waiting on the test case.

PR #35, opened 2018-03-22
Hide as much of secrets.go as we can

rthallisey:
fixes: https://github.com/automationbroker/bundle-lib/issues/8

PR #44, opened 2018-03-27
remove config package from bundle-lib

### LINKED ISSUE #43: Need to remove automationbroker/config from the dependancies
I see one use in clients/etcd.go

shawn-hurley:
Removing config package. 

Bundle lib should now take typed config structs.

fixes #43


dymurray (approved):
LGTM

djzager (approved):
LGTM. I think you like having merge conflicts (`apb/secrets.go` from #35 ).

registries/registry.go
```diff
@@ -166,24 +165,8 @@ func (r Registry) RegistryName() string {
 }
 
 // NewRegistry - Create a new registry from the registry config.
-func NewRegistry(con *config.Config, asbNamespace string) (Registry, error) {
+func NewRegistry(configuration Config, asbNamespace string) (Registry, error) {
```
djzager, line 168: I hate to do the rename thing. Just curious if we should make this `RegistryConfig` to match some of the others, like `EtcdConfig` and `ClusterConfig`.

djzager: May be better as a follow-on issue because I think this is used in more than just `registry.go`

eriknelson: @djzager This is just `Config` because it belongs to a pkg called registries, so when written out you don't end up with stutter: `registries.RegistryConfig`. It's a go convention I'm not fond of for exactly your reason, but I I've had linters yell at me for it.

eriknelson (approved):

PR #54, opened 2018-04-12
Fixes #7 - moves extract credentials to the runtime

### LINKED ISSUE #7: Move ExtractedCredentials to runtime
Move `ext_creds.go` to `runtime` package.

shawn-hurley:
* moves the methods for extracting credentials to runtime package
* adds tests for extract credentials.

This is better set up to be overriden in the future. This PR does not
allow for the ExtractCredentials to be overridden.


bundle/bind_test.go
```diff
@@ -17,43 +17,29 @@
...
-	ft "github.com/stretchr/testify/assert"
 )
 
 func TestBind(t *testing.T) {
 	t.Skip("skipping bind until we can pass in a mock client")
-	output := []byte(`
-Login failed (401 Unauthorized)
-
-PLAY [all] *********************************************************************
+	/*output := []byte(`
```
rthallisey, line 25: Any reason this is commented?

bundle/bind.go
```diff
@@ -70,12 +70,18 @@ func (e *executor) Bind(
 			}
 		}
 
-		creds, err := ExtractCredentials(
+		credBytes, err := runtime.Provider.ExtractCredentials(
```
djzager, line 73: Just want to make sure I understand this correctly. The runtime is returning `[]byte` that represent the credentials returned from the bundle and the bundle package is responsible for turning that into the `ExtractedCredentials` type.

djzager (approved):
ACK

bundle/bind_test.go
shawn-hurley: because it was not doing anything. this test was skipped.

runtime/extract_credentials.go
```diff
@@ -0,0 +1,169 @@
...
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+const (
+	// GatherCredentialsCommand - Command used when execing for bind credentials
+	// moving this constant here because eventually Extrating creds will
+	// need to be moved to runtime. Therefore keeping all of this together
+	// makes sense
+	GatherCredentialsCommand = "broker-bind-creds"
+	bundleWatchInterval      = 5
```
eriknelson, line 25: +1, I was never a fan of having this in types.

eriknelson (approved):
VISIACK, looks good.

jmrodri (approved):
ack

PR #60, opened 2018-04-23
State support for Bundles

maleck13:
Apologies on delay with this. Here is a pass at the state support in the bundle lib. Please be brutal 😀 I have changes to the broker also but will wait until this is nailed down.

@shawn-hurley You mentioned making the PR against 0.1-release. Is that needed for this work as it is targeting the next release?


bundle/executor.go
```diff
@@ -63,14 +74,17 @@ type executor struct {
 	lastStatus           StatusMessage
 	statusChan           chan StatusMessage
 	mutex                sync.Mutex
+	stateManager         StateManager
 }
 
 // NewExecutor - Creates a new Executor for running an APB.
-func NewExecutor() Executor {
+func NewExecutor(k8sClient k8.Interface, stateNS string) Executor {
```
maleck13: not sure if this is the right approach. Feedback welcome

maleck13:
Not sure how to handle version checks here as state support wont be present until the following is present https://github.com/ansibleplaybookbundle/ansible-asb-modules/pull/15 Does it mean a bump of the bundle / apb spec version and adding checks to the code for this version?

jmrodri:
Can someone point me to the genesis of this PR? IRC logs, email, or issue? Trying to get an idea what it is we're trying to solve. I think I missed or probably just forgot :) Thanks.

maleck13:
@jmrodri it is the bundle-lib part of https://github.com/openshift/ansible-service-broker/blob/master/docs/proposals/apb-state-support.md the apb state support proposal

**Notes** This is the first real thing I have done in the lib so struggled a little with understanding where the boundaries lay (ie what should be in the broker and what should be in the lib) If I am off on where the implementation of this should be, I'd be keen to know and understand more.

runtime/state.go
```diff
@@ -0,0 +1,117 @@
...
+		return false, err
+	}
+	return true, nil
+}
+
+// PrepareMount will prepare the voume definitions and mounts
+func (s *State) PrepareMount(name string) (*v1.VolumeMount, *v1.Volume, error) {
+	mount := &v1.VolumeMount{
+		Name:      name,
+		MountPath: "/etc/apb/state",
```
maleck13: set this value in PodEnv also so that it can be read by the module

bundle/bind.go
```diff
@@ -38,7 +38,7 @@ func (e *executor) Bind(
 	go func() {
 		e.actionStarted()
 		executionContext, err := e.executeApb(
-			"bind", instance.Spec, instance.Context, parameters)
+			"bind", instance, parameters)
```
shawn-hurley, line 41: 👍

bundle/executor.go
shawn-hurley: I would prefer to use the clients package to get the k8s client that way. That is how most things are behaving and I have examples of how to test that.

runtime/state.go
```diff
@@ -0,0 +1,117 @@
...
+	log "github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	kerror "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8 "k8s.io/client-go/kubernetes"
+)
+
+// State handles the state for service bundles
+type State struct {
+	Client k8.Interface
```
shawn-hurley: I would prefer this to use the `clients` package.

bundle/executor.go
```diff
@@ -141,8 +155,10 @@ func (e *executor) updateDescription(newDescription string) {
 
 // executeApb - Runs an APB Action with a provided set of inputs
 func (e *executor) executeApb(
-	action string, spec *Spec, context *Context, p *Parameters,
+	action string, instance *ServiceInstance, p *Parameters,
```
shawn-hurley, line 147: 👍

shawn-hurley (changes requested):
this is really LGTM!

I like the use of volume mounts.

bundle/executor.go
```diff
@@ -49,6 +50,16 @@ type ExecutorAsync interface {
...
 }
 
+// StateManager defines an interface for managing state created by service bundles
+type StateManager interface {
+	CopyState(fromName, toName, fromNS, toNS string) error
+	DeleteState(name string) error
+	PrepareMount(name string) (*v1.VolumeMount, *v1.Volume, error)
+	StateIsPresent(name string) (bool, error)
+	Name(instanceID string) string
+	MasterNS() string
```
jmrodri: If NS is Namespace, I'd call the method `MasterNamespace`, using ns as a variable is okay, but for the method feels better to spell it out.

bundle/bind.go
jmrodri: I'm okay with it, I prefer only to pass in what is absolutely necessary but that is just a nitpick. I also understand this makes it easier to keep the signature stable as you require more things on the instance.

bundle/bind.go
```diff
@@ -62,6 +62,12 @@ func (e *executor) Bind(
 			}
 		}
 
+		// pod execution is complete so transfer state back
+		if err := e.stateManager.CopyState(executionContext.PodName, e.stateManager.Name(instance.ID.String()), executionContext.Namespace, e.stateManager.MasterNS()); err != nil {
```
jmrodri: This is one really really long line :) can we just break it on one of the parameters?
```golang
	if err := e.stateManager.CopyState(executionContext.PodName, e.stateManager.Name(instance.ID.String()),
		executionContext.Namespace, e.stateManager.MasterNS()); err != nil {

		e.actionFinishedWithError(err)
		return
	}
```
OR
```golang
	if err := e.stateManager.CopyState(executionContext.PodName,
		e.stateManager.Name(instance.ID.String()),
		executionContext.Namespace,
		e.stateManager.MasterNS()); err != nil {

		e.actionFinishedWithError(err)
		return
	}
```

bundle/deprovision.go
```diff
@@ -44,16 +44,20 @@ func (e *executor) Deprovision(instance *ServiceInstance) <-chan StatusMessage {
...
-			executionContext.Namespace,
-			executionContext.Targets,
-			clusterConfig.Namespace,
-			clusterConfig.KeepNamespace,
-			clusterConfig.KeepNamespaceOnError,
-		)
+		executionContext, err := e.executeApb("deprovision", instance, instance.Parameters)
+		defer func() {
+			if err := e.stateManager.DeleteState(e.stateManager.Name(instance.ID.String())); err != nil {
+				log.Error("failed to delete state for instance "+instance.ID.String(), err)
```
jmrodri: I'd use the formatted version:
```golang
log.Errorf("failed to delete state for instance %s - %v", instance.ID.String(), err)
```

bundle/executor.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/pborman/uuid"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/api/core/v1"
+	k8 "k8s.io/client-go/kubernetes"
```
jmrodri: `k8` feels weird, haven't we been doing `k8s`? or `k8scli`

bundle/executor.go
```diff
@@ -214,6 +230,27 @@ func (e *executor) executeApb(
...
 	volumes, volumeMounts := buildVolumeSpecs(secrets)
+	// copy any state present for this APB over to the execution ns
+	stateName := e.stateManager.Name(instance.ID.String())
+	present, err := e.stateManager.StateIsPresent(stateName)
+	if err != nil {
+		return executionContext, err
+	}
+	if present {
+		log.Info("state: present for service instance copying to bundle namespace")
+		//copy from master ns to execution namespace
```
jmrodri: typo: space after the `//` please.

bundle/provision_or_update.go
```diff
@@ -85,6 +84,11 @@ func (e *executor) provisionOrUpdate(method executionMethod, instance *ServiceIn
 		}
 	}
 
+	// pod execution is complete so transfer state back
+	if err := e.stateManager.CopyState(executionContext.PodName, e.stateManager.Name(instance.ID.String()), executionContext.Namespace, e.stateManager.MasterNS()); err != nil {
```
jmrodri: same comment above about the super long line.

bundle/unbind.go
```diff
@@ -58,6 +57,12 @@ func (e *executor) Unbind(
 			e.actionFinishedWithError(err)
 			return
 		}
+		// pod execution is complete so transfer state back
+		if err := e.stateManager.CopyState(executionContext.PodName, e.stateManager.Name(instance.ID.String()), executionContext.Namespace, e.stateManager.MasterNS()); err != nil {
```
jmrodri: super long line comment as well

runtime/state.go
```diff
@@ -0,0 +1,117 @@
+package runtime
+
+import (
+	log "github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	kerror "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8 "k8s.io/client-go/kubernetes"
```
jmrodri: k8s?

runtime/state.go
```diff
@@ -0,0 +1,117 @@
...
+	s := &State{
+		Client:   client,
+		targetNs: targetNamespace,
+	}
+	return s
+}
+
+// CopyState copies the state configmap from one namespace to another
+func (s *State) CopyState(fromName, toName, fromNS, toNS string) error {
+	log.Info("state: copying state from namespace " + fromNS + " to ns " + toNS + " from name " + fromName + " to name " + toName)
```
jmrodri: change this to the formatted logger version:
```golang
log.Info("state: copying state from namespace %s to ns %s from name %s to name %s", fromNS, toNS, fromName, toName)
```

runtime/state.go
```diff
@@ -0,0 +1,117 @@
...
+	fromClient := s.Client.CoreV1().ConfigMaps(fromNS)
+	toClient := s.Client.CoreV1().ConfigMaps(toNS)
+	fromMap, err := fromClient.Get(fromName, metav1.GetOptions{})
+	if err != nil {
+		if kerror.IsNotFound(err) {
+			//can't copy if there is nothing to copy
```
jmrodri: space after the `//`

runtime/state.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	if _, err := toClient.Update(toMap); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Name provides a consistent name for the state object
+func (s *State) Name(id string) string {
+	return id + "-state"
```
jmrodri: this works, usually we do the `return fmt.Sprintf("%s-state", id)`

runtime/state.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// StateIsPresent checks to see is there an object carrying state for ServiceBundle
+func (s *State) StateIsPresent(stateName string) (bool, error) {
+	log.Info("state: checking is present")
+	if _, err := s.Client.CoreV1().ConfigMaps(s.targetNs).Get(stateName, metav1.GetOptions{}); err != nil {
+		if kerror.IsNotFound(err) {
+			log.Info("state: is not found ** ")
```
jmrodri: How often will this occur? Feels like this should be a debug statement not info.

runtime/state.go
```diff
@@ -0,0 +1,117 @@
...
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+// PrepareMount will prepare the voume definitions and mounts
```
jmrodri: typo in comment: voume -> volume

runtime/state.go
```diff
@@ -0,0 +1,117 @@
...
+				},
+			},
+		},
+	}
+	return mount, volume, nil
+}
+
+// DeleteState will remove the state object from the broker namespace
+func (s *State) DeleteState(name string) error {
+	log.Info("state: deleting master state " + name + " in ns " + s.targetNs)
```
jmrodri: use the formatted version of info:
```golang
log.Infof("state: deleting master state %s in ns %s", name, s.targetNs)
```

runtime/state.go
```diff
@@ -0,0 +1,117 @@
...
+	kerror "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8 "k8s.io/client-go/kubernetes"
+)
+
+// State handles the state for service bundles
+type State struct {
+	Client k8.Interface
+	// this is the namespace where the master configs will be stored
+	targetNs string
```
jmrodri: I would prefer `targetns` or `targetNS` or even better `nsTarget`

jmrodri (changes requested):
Overall decent job. Minor change requests. Thanks for the PR.

runtime/state.go
maleck13: yep left in by accident there

maleck13:
@jmrodri @shawn-hurley  ok thanks for feedback looks like we are happy with the approach. I will make the required changes and flesh out unit tests for the state module (didn't do this first as wanted to ensure I was heading in the right direction)

bundle/executor.go
rthallisey: Is ```stateNS``` different than ```bundle.serviceInstance.Context.Namespace``` https://github.com/automationbroker/bundle-lib/blob/master/bundle/types.go#L140 ?

rthallisey (commented):
Nice job.

maleck13:
@rthallisey stateNS is the where the master copy of the state will be stored. In our case the broker namespace. So it is a different namespace

runtime/state.go
maleck13: removed as we are using the clients package now

bundle/executor.go
maleck13: removed

maleck13:
addressed all requested changes. I will add a set of tests for the state manager. Note I also added a new env var to set the BUNDLE_STATE_LOCATION for use by the module

shawn-hurley:
@maleck13 So I was re-looking through this after the changes and I actually see 1 thing that should be changed. 

We are trying to move all of the stuff that uses k8scli to the runtime, and therefore I think that State Manager should be moved to the runtime package, its interface embedded into the Runtime interface and the provider struct should embed a default runtime. I also think that allowing an override in the configuration should be added. Please ping me if you have any questions.

Sorry I didn't catch this the first time around.

maleck13:
Is the idea here that a different StateManager could be provided externally when using the lib outside of the broker but we will provide a default implementation? I am ok making the change just trying to understand the reasoning. The actual concrete implementation of the interface is already in the runtime package limiting the k8scli to the runtime package. 

This makes me wonder should the state type in the runtime package be exported, seems it should be internal right?

shawn-hurley:
@maleck13 

The idea is that the only way to interact with the runtime package is through the Provider. This Provider is just an implementation of the `Runtime` interface. We should not be adding a new way to interact with runtime than through the provider. 

The idea is that if we have a bunch of sane defaults, but then if someone wants to change something they can. This will allow users of bundle lib who want to implement a new runtime can. I don't envision anyone doing this right at the moment but I do see future use cases that are not too far into the future.

maleck13:
@shawn-hurley I believe I have made the changes you requested. Let me know if this what your were expecting.

@jmrodri I believe I have addressed your feedback

runtime/runtime.go
```diff
@@ -49,6 +50,21 @@ type Configuration struct {
...
 	WatchBundle WatchRunningBundleFunc
 	ExtractedCredential
+	// StateMountLocation this is where on disk the state will be stored for a bundle
+	StateMountLocation string
+	// StateMasterNamespace the namespace where state created by bundles will be copied to between actions
+	StateMasterNamespace string
+}
+
+// StateManager defines an interface for managing state created by service bundles
+type StateManager interface {
```
shawn-hurley: I would move this interface into the state file. this would follow the pattern that we have used for ExtractedCredential

shawn-hurley (changes requested):
This a nit, but I would like to keep the same structure/layout that we have been using

shawn-hurley (approved):

runtime/state.go
```diff
@@ -0,0 +1,143 @@
+package runtime
+
+import (
+	"fmt"
+
+	"github.com/apex/log"
```
jmrodri: I think this is a mistaken import. I believe it should be log `"github.com/sirupsen/logrus"`

Remove the apex log, and add the logrus UNDER the clients, like it looks below. This makes goimports happy and uses the correct logger.

```golang
import (
    "fmt"

    "github.com/automationbroker/bundle-lib/clients"
    log "github.com/sirupsen/logrus"
    "k8s.io/api/core/v1"
    kerror "k8s.io/apimachinery/pkg/api/errors"
    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)
```

jmrodri (changes requested):
@maleck13 Wrong logger used in `state.go`, see the other comment above with the suggested fix. Then I will merge it tomorrow (Thursday). I validated it today. Great job.

maleck13:
@jmrodri  Thanks. Good spot on the import. Once merged will we tag / release I will then make the needed updates to the broker

dymurray (approved):
Looks great. This approach is going to make it much easier for APB developers to pass data between calls even outside of the cluster approach since it's just mounted files. Great job!

jmrodri (approved):

PR #63, opened 2018-04-27
Bug 1572449 - Changing type for field RHCCImage.is_official to string.

shawn-hurley:
* It appears that the json format for is_official has changed.


jwmatthews:
For background.

The stage registry showed differing behavior and changed sending booleans as strings.
The change was not intentional and has been reverted.

@shawn-hurley I think we are good to close this PR now

PR #65, opened 2018-05-01
removing the bundle execution to runtime.

shawn-hurley:
fixes #4 
* adding multiple layers of execution now.
* each executor method now controls create sandbox
* executor.executeBundle now does some stuff that is common
* runtime.RunBundle and Copy Secrets are overridable functions for the
runtime.

PR #69, opened 2018-05-02
Allow Bundle to run in a target namespace

### LINKED ISSUE #51: Bundle Lib needs to expose an option to deploy the APB in the same namespace as the target namespace
## Feature

The bundle lib at times will need to deploy the apb into the same namespace as the target of the apb. 

## Current Behavior

There is no way to currently tell bundle-lib to not create the transient namespace.

shawn-hurley:
fixes #51 

~~depends on #65~~

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -234,6 +234,15 @@ func (p provider) ValidateRuntime() error {
 	return nil
 }
 
+func isNamespaceInTargets(ns string, targets []string) bool {
+	for _, tns := range targets {
+		if tns == ns {
+			return true
+		}
+	}
+	return false
+}
+
 // CreateSandbox - Translate the broker CreateSandbox call into cluster resource calls
 func (p provider) CreateSandbox(podName string,
 	namespace string,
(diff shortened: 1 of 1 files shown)


bundle/bind.go
```diff
@@ -37,24 +44,49 @@ func (e *executor) Bind(
 
 	go func() {
 		e.actionStarted()
-		executionContext, err := e.executeApb(
-			"bind", instance.Spec, instance.Context, parameters)
+		// Create namespace name that will be used to generate a name.
+		ns := fmt.Sprintf(fmt.Sprintf("%s-%.4s-", instance.Spec.FQName, bindAction))
```
jmrodri: Why do we `fmt.Sprintf` twice?

bundle/bind.go
```diff
@@ -37,24 +44,49 @@ func (e *executor) Bind(
...
+			BundleName: pn,
+			Targets:    targets,
+			Metadata:   labels,
+			Action:     bindAction,
+			Image:      instance.Spec.Image,
+		}
+
+		serviceAccount, namespace, err := runtime.Provider.CreateSandbox(pn, ns, targets, clusterConfig.SandboxRole, labels)
+		ec.Account = serviceAccount
+		ec.Location = namespace
```
jmrodri: The order of this initialization feels weird. We create the `ExecutionContext` (partially), then get 2 values, which we add to the context. Feels like we should get the `serviceAccount` and the `namespace` first, handle the error. If it's okay, then create the ExecutionContext.

bundle/deprovision.go
```diff
@@ -42,24 +49,48 @@ func (e *executor) Deprovision(instance *ServiceInstance) <-chan StatusMessage {
 			e.actionFinishedWithError(errors.New("No image field found on instance.Spec"))
 			return
 		}
+		// Create namespace name that will be used to generate a name.
+		ns := fmt.Sprintf(fmt.Sprintf("%s-%.4s-", instance.Spec.FQName, deprovisionAction))
```
jmrodri: Double `fmt.Sprintf`?

bundle/deprovision.go
```diff
@@ -42,24 +49,48 @@ func (e *executor) Deprovision(instance *ServiceInstance) <-chan StatusMessage {
...
+			Targets:    targets,
+			Metadata:   labels,
+			Action:     deprovisionAction,
+			Image:      instance.Spec.Image,
+		}
 
-		// Might need to change up this interface to feed in instance ids
-		executionContext, err := e.executeApb("deprovision", instance.Spec,
-			instance.Context, instance.Parameters)
+		serviceAccount, namespace, err := runtime.Provider.CreateSandbox(pn, ns, targets, clusterConfig.SandboxRole, labels)
```
jmrodri: Same comment as above.

jmrodri (changes requested):
I think we should create the context after we get the service account and namespace in all of the places.

bundle/bind.go
dymurray: +1 seems redundant

bundle/bind.go
```diff
@@ -37,24 +44,49 @@ func (e *executor) Bind(
...
+			ec.BundleName,
+			ec.Location,
+			ec.Targets,
 			clusterConfig.Namespace,
 			clusterConfig.KeepNamespace,
 			clusterConfig.KeepNamespaceOnError,
 		)
 		if err != nil {
-			log.Errorf("Problem executing apb [%s] bind", executionContext.PodName)
+			log.Errorf("Problem executing apb [%s] bind", ec.BundleName)
```
dymurray: `Problem executing bundle [%s] bind` for consistency

dymurray:
The approach makes sense to me. Just a couple nits and questions Jesus mentioned but other than that looks good.

jmrodri:
@shawn-hurley needs a rebase

jmrodri (approved):
ACK

dymurray (approved):

PR #70, opened 2018-05-04
Allow dependencies to be defined on params

philipgough:
Relates to https://github.com/openshift/ansible-service-broker/issues/859

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 vendor*
 coverage-all.out
 coverage.out
+
+# IDE baggage
+.idea/
--- a/bundle/types.go
+++ b/bundle/types.go
@@ -66,11 +66,17 @@ type ParameterDescriptor struct {
 	Minimum          *NilableNumber `json:"minimum,omitempty"`
 	ExclusiveMinimum *NilableNumber `json:"exclusiveMinimum,omitempty" yaml:"exclusive_minimum,omitempty"`
 
-	Enum         []string `json:"enum,omitempty"`
-	Required     bool     `json:"required"`
-	Updatable    bool     `json:"updatable"`
-	DisplayType  string   `json:"displayType,omitempty" yaml:"display_type,omitempty"`
-	DisplayGroup string   `json:"displayGroup,omitempty" yaml:"display_group,omitempty"`
+	Enum         []string     `json:"enum,omitempty"`
+	Required     bool         `json:"required"`
+	Updatable    bool         `json:"updatable"`
+	DisplayType  string       `json:"displayType,omitempty" yaml:"display_type,omitempty"`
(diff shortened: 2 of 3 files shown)


dymurray:
I have read through the attached issue and I like the solution that you guys came up with. Since this will involve an addition to the Bundle spec this type of change will require a version bump of the Bundle Spec along with documentation updates (see [here](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/docs/developers.md#service-bundle-spec-file) for current documentation around Bundle spec).

dymurray:
From discussing at the community meeting we agreed that this will require a minor version bump of the Bundle spec bringing us to version `1.1` (per [our docs](https://github.com/ansibleplaybookbundle/ansible-playbook-bundle/blob/master/docs/developers.md#service-bundle-spec-versioning) since this is the addition of an optional field). Please update `developers.md` with an example of using this and defining its use in the Bundle spec as I linked above.

bundle/types.go
```diff
@@ -66,11 +66,17 @@ type ParameterDescriptor struct {
...
-	Required     bool     `json:"required"`
-	Updatable    bool     `json:"updatable"`
-	DisplayType  string   `json:"displayType,omitempty" yaml:"display_type,omitempty"`
-	DisplayGroup string   `json:"displayGroup,omitempty" yaml:"display_group,omitempty"`
+	Enum         []string     `json:"enum,omitempty"`
+	Required     bool         `json:"required"`
+	Updatable    bool         `json:"updatable"`
+	DisplayType  string       `json:"displayType,omitempty" yaml:"display_type,omitempty"`
+	DisplayGroup string       `json:"displayGroup,omitempty" yaml:"display_group,omitempty"`
+	Dependencies []Dependency `json:"dependencies,omitempty" yaml:"dependencies,omitempty"`
```
philipgough, line 74: We can only handle a single dependency here currently in the broker due to limitations in the frontend lib but I think to future proof where we may be able to do this if we can extend that lib, leaving this as a slice makes sense IMO. What do people think?

The spec itself should be able to support multiple dependencies if my understanding of it is correct.

cc @david-martin

dymurray:
Thanks for doing a doc PR!

.gitignore
```diff
@@ -1,3 +1,6 @@
 vendor*
 coverage-all.out
 coverage.out
+
+# IDE baggage
+.idea/
```
jmrodri, line 6: this is acceptable

bundle/types.go
jmrodri: Agreed @PhilipGough  allow multiple dependencies.

jmrodri (approved):
I'm actually okay with this, I don't see any issues.

dymurray (approved):

PR #74, opened 2018-05-08
Bug 1576881 - Add /v2/_catalog support for openshift adapter

dymurray:


dymurray:
@johnkim76 Do we have a bug for this? If so I believe we need to create 2 PRs... one against master and one against release-1.0

registries/adapters/openshift_adapter.go
```diff
@@ -51,10 +53,39 @@ func (r OpenShiftAdapter) GetImageNames() ([]string, error) {
 	log.Debug("OpenShiftAdapter::GetImageNames")
 	log.Debugf("BundleSpecLabel: %s", BundleSpecLabel)
 
-	images := r.Config.Images
-	log.Debugf("Configured to use images: %v", images)
+	if r.Config.Images != nil {
+		log.Debug("Configured to use images: %v", r.Config.Images)
```
jmrodri: Use `log.Debugf`

registries/adapters/openshift_adapter.go
```diff
@@ -51,10 +53,39 @@ func (r OpenShiftAdapter) GetImageNames() ([]string, error) {
...
+		return r.Config.Images, nil
+	}
+	log.Debug("Did not find images in config, attempting to discover from %s/v2/_catalog", r.Config.URL)
```
jmrodri: Use `log.Debugf` whenever you have a substitution string in the string. i.e. `%s` or `%v`.

registries/adapters/openshift_adapter.go
```diff
@@ -51,10 +53,39 @@ func (r OpenShiftAdapter) GetImageNames() ([]string, error) {
...
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		log.Errorf("Failed to load catalog response at %s - %v", openShiftCatalogURL, err)
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		return nil, errors.New(resp.Status)
```
jmrodri: I would add an error or warning log here. It would be nice to get as much information as to why we hit this exceptional case. I just know when we hit this we're going to wonder what really happened, and we'll add a log debug in to figure it out. If the warning or error were there already, we'd know why we got a !200.

jmrodri (dismissed):
Use *f versions of logging statements when substitution strings are used. Also add some more logging for the error conditions I noted, particularly the one where we get a !200 status code. Definitely want to print the actual status code we got to make it easier to debug in the future.

dymurray:
Thanks for the feedback @jmrodri. Made the requested changes.

shawn-hurley (approved):
VISACK

dymurray:
https://github.com/automationbroker/bundle-lib/pull/77

dymurray:
Please look through the changes on #77 and if you agree with the name change I will make the same changes here for master.

shawn-hurley (approved):

registries/adapters/partner_rhcc_adapter.go
```diff
@@ -0,0 +1,180 @@
...
+	log.Debug("PartnerRhccAdapter::GetImageNames")
+	log.Debugf("BundleSpecLabel: %s", BundleSpecLabel)
+
+	if r.Config.Images != nil {
+		log.Debugf("Configured to use images: %v", r.Config.Images)
+		return r.Config.Images, nil
+	}
+	log.Debugf("Did not find images in config, attempting to discover from %s/v2/_catalog", r.Config.URL)
+
+	req, err := http.NewRequest("GET", fmt.Sprintf(partnerCatalogURL, r.Config.URL), nil)
```
jmrodri, line 61: I think this should use the change from @maleck13 PR #87 .

jmrodri (changes requested):
I think this should use the change from @maleck13 PR #87 .

registries/adapters/partner_rhcc_adapter.go
dymurray: It looks like @maleck13's change only affects the `loadSpec` function not the Catalog call... this is unique to the partner adapter. 

I'm fine with making the change to `loadSpec` once it's merged.

jmrodri (approved):

PR #77, opened 2018-05-10
Bug 1576881 - Add discovery to openshift adapter

dymurray:


jmrodri (approved):
ACK

jwmatthews:
I have some concerns with the naming of this adapter, I think it's misleading.

The support we added in this PR was for the partner side of the RHCC registry.
It is using a combination of "Openshift Registry" for images and a custom catalog v2 endpoint they implemented themselves, it's not a fair representation of an "OpenShift Registry".

If we are considering the specific image registry implemented provided by "OpenShift Registry" it does __not__ support catalog v2 API.

There is an unrelated effort to implement an ability to list images in the actual OpenShift Registry being tracked in this card and planned for OCP 3.11:  https://trello.com/c/AZINw5qI

I think we should consider a name change of this adapter to be clearer in it's intent, perhaps something like "partner_rhcc".

jwmatthews (changes requested):
Please re-evaluate the name of this adapter, I think calling it "openshift_adapter" is misleading.

dymurray:
@jwmatthews  +1 to using `partner_rhcc`. Will make the change. Long term solution I think that we need to combine the different registry adapters which are implementing the docker registry API that is specified [here](https://docs.docker.com/registry/spec/api/) into one adapter that can handle the different schema responses so that our naming is more consistent.

dymurray:
Waiting on review of these changes before I make them to #74

jwmatthews (commented):
Name change looks good to me.

jwmatthews (approved):

registries/adapters/partner_rhcc_adapter_test.go
```diff
@@ -89,15 +89,15 @@ const AuthResponse = `
...
+func TestPartnerName(t *testing.T) {
+	partnera := PartnerRhccAdapter{}
+	ft.Equal(t, partnera.RegistryName(), "partner_rhcc", "partner_rhcc name does not match `partner_rhcc`")
 }
 
-func TestOpenShiftGetImageNames(t *testing.T) {
-	ocpa := OpenShiftAdapter{}
-	ocpa.Config.Images = Images
-	imagesFound, err := ocpa.GetImageNames()
+func TestPartnerGetImageNames(t *testing.T) {
```
eriknelson, line 97: :+1:

eriknelson (approved):
VISACK

shawn-hurley (approved):

PR #80, opened 2018-05-14
Bug 1576728 - changing remove binding

shawn-hurley:
This will allow the DAO to make better decisions about what should be stored. I think this is a hack to fix a bug because of the multiple layers of translation with multiple requests.


eriknelson (approved):
ACK, discussed with @shawn-hurley and @jmrodri. Determined this was the correct course of action for 3.10.

rthallisey (approved):

PR #81, opened 2018-05-15
ensure default bundle watcher is used if none passed in

maleck13:
Fixes the deprovision issue found here but there is a second issue with provision that I haven't got to the bottom of
https://github.com/openshift/ansible-service-broker/pull/946#issuecomment-389166217

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -135,7 +135,9 @@ func NewRuntime(config Configuration) {
 
 	defaultStateManager := state{mountLocation: config.StateMountLocation, nsTarget: config.StateMasterNamespace}
 	p := &provider{coe: cluster, ExtractedCredential: c, state: defaultStateManager}
-
+	if config.WatchBundle == nil {
+		p.watchBundle = defaultWatchRunningBundle
+	}
 	if len(config.PreCreateSandboxHooks) > 0 {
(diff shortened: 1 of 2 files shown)


maleck13:
ping @shawn-hurley

runtime/runtime.go
```diff
@@ -136,6 +136,9 @@ func NewRuntime(config Configuration) {
 	defaultStateManager := state{mountLocation: config.StateMountLocation, nsTarget: config.StateMasterNamespace}
 	p := &provider{coe: cluster, ExtractedCredential: c, state: defaultStateManager}
 
+	if config.WatchBundle == nil {
+		config.WatchBundle = defaultWatchRunningBundle
```
shawn-hurley: can we add this to the test? make sure it does not disappear again?

maleck13:
@shawn-hurley removed the deep equal as it seemed not to like function pointers. Needs more investigation. Is there anything else that should be checked manually here?

perhaps this issue https://github.com/golang/go/issues/8554

shawn-hurley:
I think that is correct, even testify cannot handle this case from my investigation.

shawn-hurley (approved):

maleck13:
closed as resolved on master

PR #82, opened 2018-05-15
Bug 1576881 - Add openshift registry adapter back in

dymurray:
We previously renamed `openshift` to `partner_rhcc` due to some confusion in the discussion around the different registries. The openshift registry is adding discovery which is dependent on https://bugzilla.redhat.com/show_bug.cgi?id=1509084 being fixed. `https://registry.connect.redhat.com` is using a proxy to approximate the catalog response so we are calling that `partner_rhcc`.

Long term we need to merge all this code into one adapter which is conformant to the docker registry API.


johnkim76:
VISIACK

mhrivnak (approved):

shawn-hurley (approved):

PR #83, opened 2018-05-17
fixed the mistake of the target image naming

jianzhangbjz:
The correct name of the image should be `registry.reg-aws.openshift.com:443/openshift3/mediawiki-apb:v3.10`, not the `openshift/openshift3/mediawiki-apb:v3.10`. Fixed bug [1579216](https://bugzilla.redhat.com/show_bug.cgi?id=1579216).

PR #85, opened 2018-05-18
Bug 1579216 - fix the target image of openshift naming

jianzhangbjz:
I cherry pick the #83, thanks all!

### DIFF
--- a/registries/adapters/openshift_adapter.go
+++ b/registries/adapters/openshift_adapter.go
@@ -27,7 +27,6 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const openShiftName = "openshift"
 const openShiftManifestURL = "%v/v2/%v/manifests/%v"
 
 // OpenShiftAdapter - Docker Hub Adapter
@@ -43,7 +42,12 @@ type OpenShiftImage struct {
 
 // RegistryName - Retrieve the registry name
 func (r OpenShiftAdapter) RegistryName() string {
-	return openShiftName
+	name := strings.Split(r.Config.URL.String(), "://")
+
+	if len(name) > 1 {
+		return name[1]
+	}
+	return name[0]
 }
 
 // GetImageNames - retrieve the images
--- a/registries/adapters/openshift_adapter_test.go
+++ b/registries/adapters/openshift_adapter_test.go
@@ -89,9 +89,30 @@ const AuthResponse = `
   "issued_at": "2018-03-27T19:54:19Z"
 }`
 
-func TestOpenShiftName(t *testing.T) {
-	ocpa := OpenShiftAdapter{}
-	ft.Equal(t, ocpa.RegistryName(), "openshift", "openshift name does not match `openshift`")
+func TestRegistryName(t *testing.T) {
+	testcases := []struct {
+		urlstr string
+		result string
+	}{
+		{"http://foo.redhat.com", "foo.redhat.com"},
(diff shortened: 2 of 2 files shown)


dymurray (dismissed):

registries/adapters/openshift_adapter.go
```diff
@@ -43,7 +42,7 @@ type OpenShiftImage struct {
 
 // RegistryName - Retrieve the registry name
 func (r OpenShiftAdapter) RegistryName() string {
-	return openShiftName
+	return strings.TrimPrefix(r.Config.URL.String(), "https://")
```
dymurray: what if I use `http://<url>` as my URL? Would prefer to see it validate the string first

jmrodri: This only removes https, if the url has any other prefix this won't work.

registries/adapters/openshift_adapter_test.go
```diff
@@ -28,6 +28,8 @@ import (
 )
 
 var Images = []string{"satoshi/nakamoto", "foo/bar", "paul/atreides"}
+var TestURL = "https://registry.reg-aws.openshift.com:443"
+var RegistryName = "registry.reg-aws.openshift.com:443"
```
jmrodri: No need to declare these strings here. I would localize them to the test case instead.

dymurray:
https://github.com/automationbroker/bundle-lib/pull/86

Submitted a PR to master to attempt to address `http://`

registries/adapters/openshift_adapter_test.go
```diff
@@ -89,9 +91,17 @@ const AuthResponse = `
...
 }`
 
-func TestOpenShiftName(t *testing.T) {
-	ocpa := OpenShiftAdapter{}
-	ft.Equal(t, ocpa.RegistryName(), "openshift", "openshift name does not match `openshift`")
+func TestRegistryName(t *testing.T) {
+	url, err := url.Parse(TestURL)
+	if err != nil {
+		t.Fatal("Error: ", err)
+	}
```
jmrodri: Consider using a table for the test cases, this will allow us to easily add new urls if we want to test other conditions:
```golang
func TestRegistryName(t *testing.T) {

    testcases := []struct {
        urlstr string
        result string
    }{
        {"http://foo.redhat.com", "foo.redhat.com"},
        {"https://registry.reg-aws.openshift.com:443", "registry.reg-aws.openshift.com:443"},
    }

    for _, testcase := range testcases {
        url, err := url.Parse(testcase)
        if err != nil {
            t.Fatal("Error: ", err)
        }

        ocpa := OpenShiftAdapter{
            Config: Configuration{
                URL: url,
            },
        }
        ft.Equal(t, ocpa.RegistryName(), result, "the registry name returned did not match expected value")
    }
}
```

jmrodri (changes requested):
Thanks for the contribution. A few changes will make it perfect. We'll want to strip probably all the prefixes so we will probably need to find the index of the `//` and then strip everything before that index. And using a table in the test case will allow more urls to be added easily.

jianzhangbjz:
@dymurray @jmrodri Many thanks for your great feedback. I updated them according to your comments and added the case conversion, please have a review.

dymurray (approved):

jianzhangbjz:
@dymurray @jmrodri I removed the case conversion as said in #86.

jmrodri (approved):

PR #86, opened 2018-05-18
Update openshift adapter and tests for http

dymurray:


registries/adapters/openshift_adapter_test.go
```diff
@@ -96,12 +97,23 @@ func TestRegistryName(t *testing.T) {
...
 		},
 	}
+	ocpaHttp := OpenShiftAdapter{
+		Config: Configuration{
+			URL: urlHttp,
+		},
+	}
 	ft.Equal(t, ocpa.RegistryName(), RegistryName, "the registry name returned did not match expected value")
+	ft.Equal(t, ocpaHttp.RegistryName(), RegistryName, "the registry name returned did not match expected value")
 }
```
jmrodri: Consider making this a table driven testcase.

```golang
func TestRegistryName(t *testing.T) {

    testcases := []struct {
        urlstr string
        result string
    }{
        {"http://foo.redhat.com", "foo.redhat.com"},
        {"https://registry.reg-aws.openshift.com:443", "registry.reg-aws.openshift.com:443"},
        {"ftp://registry.reg-aws.openshift.com", "registry.reg-aws.openshift.com"},
        {"file:///home/user/awesome", "/home/user/awesome"},
    }

    for _, testcase := range testcases {
        url, err := url.Parse(testcase)
        if err != nil {
            t.Fatal("Error: ", err)
        }

        ocpa := OpenShiftAdapter{
            Config: Configuration{
                URL: url,
            },
        }
        ft.Equal(t, ocpa.RegistryName(), result, "the registry name returned did not match expected value")
    }
}
```

registries/adapters/openshift_adapter_test.go
```diff
@@ -29,6 +29,7 @@ import (
 
 var Images = []string{"satoshi/nakamoto", "foo/bar", "paul/atreides"}
 var TestURL = "https://registry.reg-aws.openshift.com:443"
+var TestHttpURL = "http://registry.reg-aws.openshift.com:443"
 var RegistryName = "registry.reg-aws.openshift.com:443"
```
jmrodri: remove TestURL, TestHttpURL, and RegistryName, just define them in the testcase directly.

registries/adapters/openshift_adapter.go
```diff
@@ -42,7 +42,11 @@ type OpenShiftImage struct {
 
 // RegistryName - Retrieve the registry name
 func (r OpenShiftAdapter) RegistryName() string {
-	return strings.TrimPrefix(r.Config.URL.String(), "https://")
+	var name string
+	name = strings.TrimPrefix(r.Config.URL.String(), "https://")
+	name = strings.TrimPrefix(name, "http://")
+
+	return name
```
jmrodri: I'm going to assume this url can be ANY valid url. If that's the case I'd rather strip off any leading `://` that comes before it. So I'd rather do something like this:

```golang
func (r OpenShiftAdapter) RegistryName() string {
	name := strings.Split(r.Config.URL.String(), "://")

	if len(name) > 1 {
		return name[1]
	}
	return name[0]
```

This will cover all URLs `file://`, `http://`, `ftp://`, `https://`.

jmrodri (changes requested):
NACK

registries/adapters/openshift_adapter_test.go
dymurray: Thanks that does look much better. Appreciate the feedback.

dymurray:
Thanks @jmrodri made the requested changes. Great feedback.

dymurray:
Thanks @jmrodri made the requested changes. Great feedback.

registries/adapters/openshift_adapter.go
```diff
@@ -42,7 +42,12 @@ type OpenShiftImage struct {
 
 // RegistryName - Retrieve the registry name
 func (r OpenShiftAdapter) RegistryName() string {
-	return strings.TrimPrefix(r.Config.URL.String(), "https://")
+	name := strings.Split(r.Config.URL.String(), "://")
```
jianzhangbjz, line 45: Do we need to add the case conversion in here since the `url` ccannot cover the Uppercase? Like: `"http://FOO.REDHAT.COM"` should be chaned to `foo.redhat.com` after this `func`.

jmrodri: No need to do this. The value of `r.Config.URL` is a `*url.URL`. So it would need to be a valid url BEFORE that was ever set. We go from string to URL early in the startup process. The parsing of the http://FOO.REDHAT.COM would fail before the broker starts up. So we're okay here.

jmrodri (approved):

registries/adapters/openshift_adapter.go
jianzhangbjz: Thanks, I got it and also updated #85.

registries/adapters/openshift_adapter_test.go
```diff
@@ -92,16 +90,29 @@ const AuthResponse = `
...
 
 func TestRegistryName(t *testing.T) {
-	url, err := url.Parse(TestURL)
-	if err != nil {
-		t.Fatal("Error: ", err)
+	testcases := []struct {
+		urlstr string
+		result string
+	}{
+		{"http://foo.redhat.com", "foo.redhat.com"},
```
eriknelson, line 97: :+1:

eriknelson (approved):

PR #87, opened 2018-05-18
fix issue with runtime label

maleck13:
As discussed in the following PR  https://github.com/openshift/ansible-service-broker/pull/946 a change to the label to use "bundle" instead of "apb" caused an issue that would cause the broker to assume the bundle was using runtime=1 rather than runtime=2 this manifested itself as an attempted exec into the container.

- update the runtime test to check for the watchBundle func
- update adapter to deal with old and new runtime labels (minor refactor to allow testing)
- add some parsing tests

### DIFF
--- a/registries/adapters/adapter.go
+++ b/registries/adapters/adapter.go
@@ -20,14 +20,15 @@ import (
 	"bytes"
 	b64 "encoding/base64"
 	"encoding/json"
-	"io/ioutil"
-	"net/http"
 	"net/url"
 	"strconv"
 
+	"io/ioutil"
+	"net/http"
+
 	"github.com/automationbroker/bundle-lib/bundle"
 	log "github.com/sirupsen/logrus"
-	yaml "gopkg.in/yaml.v1"
+	"gopkg.in/yaml.v1"
 )
 
 // Adapter - Adapter will wrap the methods that a registry needs to fully manage images.
@@ -57,21 +58,32 @@ type Configuration struct {
 	Tag        string
 }
 
-// Retrieve the spec from a registry manifest request
-func imageToSpec(req *http.Request, image string) (*bundle.Spec, error) {
-	log.Debug("Registry::imageToSpec")
-	spec := &bundle.Spec{}
-	req.Header.Add("Accept", "application/json")
-
-	resp, err := http.DefaultClient.Do(req)
+func commonResponseHandler(resp *http.Response) ([]byte, error) {
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return nil, err
 	}
-	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusUnauthorized {
+		log.Errorf("Unable to authenticate to the registry, registry credentials could be invalid.")
+		return nil, nil
+	}
 
+	if resp.StatusCode != http.StatusOK {
+		log.Errorf("unexpected response code %v body %v", resp.StatusCode, string(body))
+		return nil, nil
+	}
+	return body, nil
+}
+
+// Retrieve the spec from a registry manifest request
+func imageToSpec(imageDetails []byte, image string) (*bundle.Spec, error) {
+	log.Debug("Registry::imageToSpec")
+	spec := &bundle.Spec{}
 	type label struct {
-		Spec    string `json:"com.redhat.apb.spec"`
-		Runtime string `json:"com.redhat.bundle.runtime"`
+		Spec             string `json:"com.redhat.apb.spec"`
+		LegacyAPBRuntime string `json:"com.redhat.apb.runtime"`
+		BundleRuntime    string `json:"com.redhat.bundle.runtime"`
 	}
 
 	type config struct {
@@ -86,24 +98,8 @@ func imageToSpec(req *http.Request, image string) (*bundle.Spec, error) {
 		Config *config `json:"config"`
 	}{}
 
-	if resp.StatusCode == http.StatusUnauthorized {
-		log.Errorf("Unable to authenticate to the registry, registry credentials could be invalid.")
-		return nil, nil
-	}
-
-	// resp.Body is an io.Reader, which are a one time use.  Save the
-	// contents to a byte[] for debugging, then remake the io.Reader for the
-	// JSON decoder.
-	debug, _ := ioutil.ReadAll(resp.Body)
-	if resp.StatusCode != http.StatusOK {
-		log.Errorf("Image '%s' may not exist in registry.", image)
-		log.Error(string(debug))
-		return nil, nil
-	}
-
-	r := bytes.NewReader(debug)
-	err = json.NewDecoder(r).Decode(&hist)
-	if err != nil {
+	r := bytes.NewReader(imageDetails)
+	if err := json.NewDecoder(r).Decode(&hist); err != nil {
 		log.Errorf("Error grabbing JSON body from response: %s", err)
 		return nil, err
 	}
@@ -113,8 +109,7 @@ func imageToSpec(req *http.Request, image string) (*bundle.Spec, error) {
 		return nil, nil
 	}
 
-	err = json.Unmarshal([]byte(hist.History[0]["v1Compatibility"]), &conf)
-	if err != nil {
+	if err := json.Unmarshal([]byte(hist.History[0]["v1Compatibility"]), &conf); err != nil {
 		log.Errorf("Error unmarshalling intermediary JSON response: %s", err)
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


registries/adapters/adapter.go
```diff
@@ -20,14 +20,15 @@ import (
...
 	b64 "encoding/base64"
 	"encoding/json"
-	"io/ioutil"
-	"net/http"
 	"net/url"
 	"strconv"
 
+	"io/ioutil"
+	"net/http"
+
```
jmrodri: Not sure what caused these to be moved. `goimports` keeps these stdlibs in alphabetical order with the others. I'd move these 2 back to where they were.

registries/adapters/adapter.go
```diff
@@ -20,14 +20,15 @@ import (
...
 	"github.com/automationbroker/bundle-lib/bundle"
 	log "github.com/sirupsen/logrus"
-	yaml "gopkg.in/yaml.v1"
+	"gopkg.in/yaml.v1"
```
jmrodri: Any reason you removed the prefix? It works either way, but the first one wasn't necessarily wrong.

registries/adapters/adapter.go
```diff
@@ -57,21 +58,32 @@ type Configuration struct {
...
-	defer resp.Body.Close()
+	if resp.StatusCode == http.StatusUnauthorized {
+		log.Errorf("Unable to authenticate to the registry, registry credentials could be invalid.")
+		return nil, nil
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		log.Errorf("unexpected response code %v body %v", resp.StatusCode, string(body))
+		return nil, nil
+	}
```
jmrodri: So `!StatusOK` and `StatusUnauthorized` are *NOT* error conditions? seems like these should return some sort of error. i.e. `return nil, fmt.Error("Unauthorized")` vs nil.

registries/adapters/adapter.go
```diff
@@ -57,21 +58,32 @@ type Configuration struct {
...
+	return body, nil
+}
 
+// Retrieve the spec from a registry manifest request
+func imageToSpec(imageDetails []byte, image string) (*bundle.Spec, error) {
```
jmrodri, line 88: This makes way more sense. The method simply converts data not how to get the data. :+1:

registries/adapters/adapter.go
```diff
@@ -113,8 +109,7 @@ func imageToSpec(req *http.Request, image string) (*bundle.Spec, error) {
 		return nil, nil
 	}
 
-	err = json.Unmarshal([]byte(hist.History[0]["v1Compatibility"]), &conf)
-	if err != nil {
+	if err := json.Unmarshal([]byte(hist.History[0]["v1Compatibility"]), &conf); err != nil {
```
jmrodri, line 120: much more concise, I like this paradigm better. :+1:

registries/adapters/adapter.go
```diff
@@ -139,7 +134,14 @@ func imageToSpec(req *http.Request, image string) (*bundle.Spec, error) {
...
 
-	spec.Runtime, err = getAPBRuntimeVersion(conf.Config.Label.Runtime)
+	version := conf.Config.Label.LegacyAPBRuntime
+
+	// prefer bundle runtime
+	if conf.Config.Label.BundleRuntime != "" {
+		log.Debugf("bundle runtime present using this over apb runtime")
+		version = conf.Config.Label.BundleRuntime
+	}
+	spec.Runtime, err = getAPBRuntimeVersion(version)
```
jmrodri, line 152: Excellent! using a default then override. :+1:

registries/adapters/adapter_test.go
```diff
@@ -17,11 +17,90 @@
...
 
+	"github.com/automationbroker/bundle-lib/bundle"
 	ft "github.com/stretchr/testify/assert"
 )
 
 func TestSpecLabel(t *testing.T) {
 	ft.Equal(t, BundleSpecLabel, "com.redhat.apb.spec", "spec label does not match dockerhub")
 }
+
+func TestImageToSpec(t *testing.T) {
```
jmrodri, line 33: YES! a new test. I love that you are writing tests. It's excellent.

registries/adapters/adapter_test.go
```diff
@@ -17,11 +17,90 @@
...
+	cases := []struct {
+		Name     string
+		Response history
+		Validate func(t *testing.T, spec *bundle.Spec)
+	}{
+		{
+			Name: "test spec parsed correctly and runtime version 1 and spec version 1.0 when no Label present",
+			Response: history{History: []map[string]string{
+				{
+					"v1Compatibility": "{\"config\":{\"Labels\":{\"build-date\":\"20170801\",\"com.redhat.apb.spec\":\"" + testApbSpec + "\",\"com.redhat.apb.version\":\"0.1.0\"}}}",
```
jmrodri: I'd probably use `fmt.Sprintf`

```golang
"v1Compatibility": fmt.Sprintf("{\"config\":{\"Labels\":{\"build-date\":\"20170801\",\"com.redhat.apb.spec\":\"%s\",\"com.redhat.apb.version\":\"0.1.0\"}}}", testApbSpec)
```

Do that for all three of the strings below.

registries/adapters/dockerhub_adapter.go
```diff
@@ -231,7 +231,14 @@ func (r DockerHubAdapter) loadSpec(imageName string) (*bundle.Spec, error) {
 		return nil, err
 	}
 	req.Header.Set("Authorization", fmt.Sprintf("Bearer %v", token))
-	return imageToSpec(req, fmt.Sprintf("%s/%s:%s", r.RegistryName(), imageName, r.Config.Tag))
+	req.Header.Add("Accept", "application/json")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	body, err := commonResponseHandler(resp)
```
jmrodri: What do we do with the `err`? looks like we're dropping it on the floor.

registries/adapters/openshift_adapter.go
```diff
@@ -145,5 +145,14 @@ func (r OpenShiftAdapter) loadSpec(imageName string) (*bundle.Spec, error) {
...
 	req.Header.Set("Authorization", fmt.Sprintf("Bearer %v", token))
-	return imageToSpec(req, fmt.Sprintf("%s/%s:%s", r.RegistryName(), imageName, r.Config.Tag))
+	req.Header.Add("Accept", "application/json")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	body, err := commonResponseHandler(resp)
+	if err != nil {
+		return nil, fmt.Errorf("OpenShiftAdapter::error handling openshift registery response %s", err)
```
jmrodri: Should this be `%v` for the `err` or is `%s` okay? I guess it's okay.

jmrodri (changes requested):
Some changes, but I like the approach. Nice job.

registries/adapters/adapter.go
```diff
@@ -57,21 +58,32 @@ type Configuration struct {
...
 
+// Retrieve the spec from a registry manifest request
+func imageToSpec(imageDetails []byte, image string) (*bundle.Spec, error) {
+	log.Debug("Registry::imageToSpec")
+	spec := &bundle.Spec{}
 	type label struct {
-		Spec    string `json:"com.redhat.apb.spec"`
-		Runtime string `json:"com.redhat.bundle.runtime"`
+		Spec             string `json:"com.redhat.apb.spec"`
+		LegacyAPBRuntime string `json:"com.redhat.apb.runtime"`
```
eriknelson, line 93: Ah, I see how this snuck in. Good catch.

eriknelson (approved):
LGTM

registries/adapters/adapter.go
eriknelson: Had the same question independently. AFAICT these were refactored from previous code. I'm wondering if maybe we should return errors here, and the calling code can choose to care about it.

dymurray: +1 I would prefer to see an error here especially on an Unauthorized. This way as we begin to refactor to the adapters we can have some sanity checks using the `commonResponseHandler` with grabbing auth tokens.

dymurray:
I'm indifferent to some of the nits but I agree with @jmrodri comments about missing some error handling. Other than that it looks good.

registries/adapters/adapter.go
maleck13: Yeah changed my editor so might be some settings I need to tweak

registries/adapters/adapter.go
maleck13: No again I think this was the new editor

registries/adapters/adapter.go
maleck13: As @eriknelson mentioned, this code was just moved it functions as it did before. I am happy to update to the suggested approach though. Going to rename the method to ```registryResponseHandler``` also

registries/adapters/dockerhub_adapter.go
maleck13: Good spot this was a miss. Perhaps we should look at using errcheck as part of our make check? I can open an issue. It checks that all errors are being handled

registries/adapters/openshift_adapter.go
maleck13: yes it just takes the Error method response and appends it https://play.golang.org/p/O0YzY_43Awl

maleck13:
errcheck issue created https://github.com/automationbroker/bundle-lib/issues/89
PR updated

dymurray (approved):

jmrodri (approved):

PR #88, opened 2018-05-19
add prune rules

jmrodri:
The prune rules will reduce the size of the vendor directory:

BEFORE:
```
$ du -sh vendor/
244M    vendor/
```

AFTER:
```
$ du -sh vendor/
32M     vendor/
```

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -57,5 +57,5 @@
   version = "1.3.1"
 
 [prune]
-  unused-packages = true
   go-tests = true
+  non-go = true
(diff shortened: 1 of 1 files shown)


shawn-hurley:
I am wondering if some developers might want to have the entire graph in vendor? A consumer of bundle lib can choose to implement prune and that will prune the dependencies that bundle-lib does not need for them. This is the nice part of not checking in vendor to bundle lib IMO.

Think about the case that I add a new dependency. I would need to first add the import path to my go file, and then I would need to run dep ensure for it to be "un-pruned". Thoughts?

jmrodri:
@shawn-hurley removed unused-packages and added non-go. That sounds like a valid developer experience use case. Vendor does grow from 32M back to 76M but that is a far cry from 244M.

shawn-hurley (approved):

eriknelson (approved):

PR #93, opened 2018-05-24
state: use correct name for configmap mount add back env var

maleck13:
There was a recent change in the bundle lib that changed the name of the configmap to mount. This caused issues with the state support where the pod would fail to start if there is state found as it would attempt to mount a configmap that didn't exist
Also the env var for the state location was removed. This is used by the module currently. 

Resolves issues found in latest 0.2 branch https://github.com/openshift/ansible-service-broker/pull/946

**changes**
 - use correct configmap name in the mount
 - set the env var
 - change the name of the method that returns the master state configmap name (ie the state configmap stored in the broker NS) to be clearer

### DIFF
--- a/runtime/run_bundle_test.go
+++ b/runtime/run_bundle_test.go
@@ -6,13 +6,26 @@ import (
 	"strings"
 	"testing"
 
+	"bytes"
+	"io/ioutil"
+	"net/http"
+
 	"github.com/automationbroker/bundle-lib/clients"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/kubernetes/scheme"
+	fakerest "k8s.io/client-go/rest/fake"
 )
 
 func TestDefaultRunBundle(t *testing.T) {
+	restClient := &fakerest.RESTClient{
+		Resp: &http.Response{
+			StatusCode: http.StatusOK,
+			Body:       ioutil.NopCloser(bytes.NewReader([]byte(`{"major":"3", "minor": "2"}`))),
+		},
+		NegotiatedSerializer: scheme.Codecs,
+	}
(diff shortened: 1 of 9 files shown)


runtime/runtime.go
```diff
@@ -58,8 +58,19 @@ type Configuration struct {
 	StateMountLocation string
 	// StateMasterNamespace the namespace where state created by bundles will be copied to between actions
 	StateMasterNamespace string
+	// Platform indicates which platform the runtime is for
+	Platform platform
```
shawn-hurley: Sorry I don't know the use case for this can you explain?

shawn-hurley (changes requested):

runtime/runtime.go
maleck13: I will remove as probably shouldn't be part of this PR as mentioned in the initial comment I wasn't sure if this was something wanted. 
The reason it came about was I noticed when trying to update the tests that I needed to initialize the runtime provider. However this now tries to call out to OpenShift to figure out what type of cluster / platform it is running in. I added this as I thought the caller of bundle should be in a position to know / specify where they are running rather than the lib deciding and would mean we could remove the callout in the NewRuntime method?

runtime/run_bundle_test.go
```diff
@@ -91,6 +105,86 @@ func TestDefaultRunBundle(t *testing.T) {
...
+				}
+				if pod.Spec.Volumes[0].ConfigMap == nil {
+					t.Fatalf("expected the volume to be a configmap but was nil")
+				}
+				if pod.Spec.Volumes[0].ConfigMap.Name != pod.Name {
+					t.Fatalf("expected the configmap name to match the pod name but it was %s ", pod.Spec.Volumes[0].ConfigMap.Name)
+				}
+
+			},
+
```
shawn-hurley: NIT: Can we remove the extra space?

shawn-hurley (approved):

jmrodri (approved):
LGTM

Issue #89, opened 2018-05-21
add errcheck as part of the make check command


maleck13:
Sometimes it is possible to miss handling an error. Rather than relying on review to catch this we could include the popular errcheck tool as part of the ```make check``` command.

https://github.com/kisielk/errcheck

Happy to take this on if we want to do it

eriknelson:
+1, this sounds useful!

jmrodri:
We decided against implementing this

> After some thought I don't think this should be merged. Wrapping the `Close()` in a function simply to log the error to satisfy the tool smells, since we are not actually "handling" the error. 
>
> Also, having the errcheck tool doesn't really force me to handle the error other than possibly logging it which isn't really handling it.
>
> And finally I would like to be able to ignore errors, sometimes I just don't care. resp.Body.Close() is precisely one of those situations where I don't care.
>
> I brought this up with @eriknelson and @shawn-hurley, so we're going to close this PR.

PR #95, opened 2018-05-25
add errcheck support

maleck13:
Resolves issue https://github.com/automationbroker/bundle-lib/issues/89

This pr adds the errcheck tool that scans for unchecked errs in the code. It will run be default as part of the travis build and also if you run ```make check```

### DIFF
--- a/registries/adapters/dockerhub_adapter.go
+++ b/registries/adapters/dockerhub_adapter.go
@@ -149,7 +149,11 @@ func (r DockerHubAdapter) getDockerHubToken() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			log.Warn("failed to close response body : %s", err)
+		}
+	}()
 
 	jsonToken, err := ioutil.ReadAll(resp.Body)
 
@@ -183,7 +187,11 @@ func (r DockerHubAdapter) getNextImages(ctx context.Context,
 		close(ch)
 		return nil, err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			log.Warn("failed to close response body : %s", err)
+		}
+	}()
 
 	imageList, err := ioutil.ReadAll(resp.Body)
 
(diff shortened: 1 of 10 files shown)


maleck13:
investigating build failure

registries/adapters/dockerhub_adapter.go
```diff
@@ -265,12 +277,15 @@ func (r DockerHubAdapter) getBearerToken(imageName string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer response.Body.Close()
+	defer func() {
```
shawn-hurley, line 280: Are we doing this because we added the error check to the `check` command in the makefile?

I actually don't like adding these go routines just because of a tool. Wondering we can just add this to the make file, and not add it to CI. The tool is useful for checking the code, but I don't know if we want to add this code when we really do want to ignore the error.

shawn-hurley (changes requested):

shawn-hurley:
We also need to add the go get to the travis job. That is what is failing the build.

registries/adapters/dockerhub_adapter.go
maleck13: When you say adding go routines, I am not sure I follow do you mean the additional function call as we are really are just deferring one function call over another. Originally we were deferring the Close method where as now we are deferring an anonymous function that calls the ```Close``` method. There are no additional  go routines created?

I know where you are coming from with ignoring the error when closing the response body, but do we actually want to ignore it? What is the downside to not ignoring it?
I do think there is value in having the err check tool run as part of the PR process as it gives us the confidence that any new code has at least handled any possible errors. I think a couple of additional err checks as a cost is not that high

maleck13:
@shawn-hurley I did add it to the travis file https://github.com/automationbroker/bundle-lib/pull/95/files#diff-354f30a63fb0907d4ad57269548329e3 . Is that the wrong location?

shawn-hurley (approved):

registries/adapters/dockerhub_adapter.go
jmrodri: the defer func() solely for errcheck is annoying because you really don't need to check the err on a close() it just gets ignored. What are we going to do if the response body doesn't get closed? If this ends up logging a bunch of warnings to the broker log because of errors that we would've ignored before I will ask we remove these. For now, I'll let it stand.

jmrodri (approved):
I need the indifferent approval button :) that is I'll let this change go in but I don't think it's really necessary.

jmrodri (changes requested):
After some thought I don't think this should be merged. Wrapping the `Close()` in a function simply to log the error to satisfy the tool smells, since we are not actually "handling" the error. 

Also, having the errcheck tool doesn't really force me to handle the error other than possibly logging it which isn't really handling it.

And finally I would like to be able to ignore errors, sometimes I just don't care. resp.Body.Close() is precisely one of those situations where I don't care.

I brought this up with @eriknelson and @shawn-hurley, so we're going to close this PR.

PR #98, opened 2018-05-30
adds oauth2 support to the rhcc adapter

mhrivnak:
also adds config option "skip_verify_ssl" to the rhcc adapter so that the oauth
feature can be tested against internal registries.

### DIFF
--- a/registries/adapters/oauth/client.go
+++ b/registries/adapters/oauth/client.go
@@ -1,0 +1,229 @@
+//
+// Copyright (c) 2018 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package oauth
+
+import (
+	"crypto/tls"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"regexp"
+	"sync"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// authResponse - holds the response data from an oauth2 token request
+type authResponse struct {
+	Token string `json:"access_token"`
+}
+
+// NewClient - creates and returns a *Client ready to use. If skipVerify is
+// true, it will skip verification of the remote TLS certificate.
+func NewClient(user, pass string, skipVerify bool, url *url.URL) *Client {
+	transport := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: skipVerify}}
+	if skipVerify == true {
+		log.Warn("skipping verification of registry TLS certificate per adapter configuration")
+	}
+
+	return &Client{
+		user:   user,
+		pass:   pass,
+		url:    url,
+		mutex:  &sync.Mutex{},
+		client: &http.Client{Timeout: time.Second * 60, Transport: transport},
+	}
+}
+
+// Client - may be used as an HTTP client that automatically authenticates using oauth.
+type Client struct {
+	user   string
+	pass   string
+	token  string
+	mutex  *sync.Mutex
+	client *http.Client
+	url    *url.URL
+}
+
+// NewRequest - creates and returns a *http.Request assuming the GET method.
+// The base URL configured on the Client gets used with its Path component
+// replaced by the path argument. If a token is available, it is added to the
+// request automatically. "Accept: application/json" is added to all requests.
+// The caller should customize the request as necessary before using it.
+func (c *Client) NewRequest(path string) (*http.Request, error) {
+	req, err := http.NewRequest("GET", c.url.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+	req.URL.Path = path
+	if c.token != "" {
+		req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", c.token))
+	}
+	req.Header.Add("Accept", "application/json")
+	return req, nil
+}
+
+// Do - passes through to the underlying http.Client instance
+func (c *Client) Do(req *http.Request) (*http.Response, error) {
+	return c.client.Do(req)
+}
+
+// Getv2 - makes a GET request to the registry's /v2/ endpoint. If a 401
+// Unauthorized response is received, this method attempts to obtain an oauth
+// token and tries again with the new token. If a username and password are
+// available, they are used with Basic Auth in the request to the token
+// service. This method is goroutine-safe.
+func (c *Client) Getv2() error {
+	// lock to prevent multiple goroutines from retrieving, and especially
+	// writing, a new token at the same time
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	req, err := c.NewRequest("/v2/")
+	if err != nil {
+		return err
+	}
+	resp, err := c.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
+	case http.StatusUnauthorized:
+		h := resp.Header.Get("www-authenticate")
+		c.getToken(h)
+
+		// try the new token
+		tokenReq, err := c.NewRequest("/v2/")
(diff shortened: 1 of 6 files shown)


mhrivnak:
There is a companion PR to the broker that will surface the tls setting, but I can't make it until the broker gets updated with a more recent version of bundle-lib. For now you can get a preview here:

https://github.com/openshift/ansible-service-broker/compare/bundle-lib-0.2...mhrivnak:skip-verify?expand=1

registries/adapters/oauth/client.go
```diff
@@ -0,0 +1,229 @@
...
+		log.Debug("GET /v2/ successful with new token")
+
+	case http.StatusOK:
+		if c.token == "" {
+			log.Debug("GET /v2/ successful without token")
+		} else {
+			log.Debug("GET /v2/ successful with existing token")
+		}
+
+	default:
```
dymurray, line 139: +1

registries/adapters/oauth/client.go
```diff
@@ -0,0 +1,229 @@
...
+		return nil, err
+	}
+	rservice, err := regexp.Compile("service=\"([^\"]+)\"")
+	if err != nil {
+		return nil, err
+	}
+
+	rmatch := rrealm.FindStringSubmatch(value)
+	smatch := rservice.FindStringSubmatch(value)
+	if len(rmatch) != 2 || len(smatch) != 2 {
```
dymurray: If there is no `service` defined in the www-auth then this will fail. 

Try: 
```
curl -k https://registry.reg-aws.openshift.com/v2/ -vvv
```

dymurray (changes requested):

registries/adapters/oauth/client.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	// make the request
+	resp, err := c.client.Do(req)
+	if err != nil {
+		msg := fmt.Sprintf("error obtaining token: %s", err.Error())
+		log.Warn(msg)
+		return err
+	}
+	defer resp.Body.Close()
```
shawn-hurley, line 175: You may need to follow the code that is coming in for err check here. 

I don't want to have to make that PR make the change here if we want to bring this in.

registries/adapters/oauth/client.go
```diff
@@ -0,0 +1,233 @@
...
+	realm := rmatch[1]
+
+	u, err := url.Parse(realm)
+	if err != nil {
+		msg := fmt.Sprintf("realm is not a valid URL: %s", realm)
+		log.Warn(msg)
+		return nil, errors.New(msg)
+	}
+
+	smatch := rservice.FindStringSubmatch(value)
```
dymurray, line 224: Awesome thanks for fixing.

registries/adapters/oauth/client_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"strings"
+	"testing"
+)
+
+var headerCases = map[string]string{
+	"Bearer realm=\"http://foo/a/b/c\",service=\"bar\"":  "http://foo/a/b/c?service=bar",
+	"Bearer service=\"bar\",realm=\"http://foo/a/b/c\"":  "http://foo/a/b/c?service=bar",
+	"Bearer realm=\"http://foo/a/b/c/\",service=\"bar\"": "http://foo/a/b/c/?service=bar",
+	"Bearer realm=\"https://foo\",service=\"bar\"":       "https://foo?service=bar",
+	"Bearer realm=\"http://foo/a/b/c\"":                  "http://foo/a/b/c",
```
dymurray, line 30: And awesome tests!

registries/adapters/rhcc_adapter.go
```diff
@@ -123,13 +133,12 @@ func (r RHCCAdapter) loadSpec(imageName string) (*bundle.Spec, error) {
 	if r.Config.Tag == "" {
 		r.Config.Tag = "latest"
 	}
-	req, err := http.NewRequest("GET",
-		fmt.Sprintf("%v/v2/%v/manifests/%v", r.Config.URL.String(), imageName, r.Config.Tag), nil)
+	req, err := r.client.NewRequest(fmt.Sprintf("/v2/%v/manifests/%v", imageName, r.Config.Tag))
```
dymurray, line 138: I like this approach so much better.

dymurray (approved):
Great job!

registries/adapters/oauth/client.go
```diff
@@ -0,0 +1,233 @@
...
+	if len(smatch) == 2 {
+		service := smatch[1]
+		q := u.Query()
```
shawn-hurley, line 227: NIT: `u.Query().Set("service", service)` IMO is cleaner

registries/adapters/rhcc_adapter.go
```diff
@@ -88,16 +98,16 @@ func (r RHCCAdapter) FetchSpecs(imageNames []string) ([]*bundle.Spec, error) {
...
+func (r RHCCAdapter) loadImages(query string) (RHCCImageResponse, error) {
 	log.Debug("RHCCRegistry::LoadImages")
-	log.Debug("Using " + r.Config.URL.String() + " to source APB images using query:" + Query)
-	req, err := http.NewRequest("GET",
-		fmt.Sprintf("%v/v1/search?q=%v", r.Config.URL.String(), Query), nil)
+	req, err := r.client.NewRequest("/v1/search")
 	if err != nil {
 		return RHCCImageResponse{}, err
 	}
+	req.URL.RawQuery = fmt.Sprintf("q=%s", query)
```
shawn-hurley: Do we maybe want to use [url.Values](https://godoc.org/net/url#Values) and encode the values using this?
I only bring it up because we are using it above and maybe want to use it here?

shawn-hurley (changes requested):
nits and a question

registries/adapters/oauth/client.go
mhrivnak: I've read this a bunch of times, and I honestly can't figure out what you're trying to tell me. :) Can you elaborate and/or rephrase?

I'm confused because I'm checking the http response codes below, and I'm not sure what other PR you're talking about.

registries/adapters/oauth/client.go
mhrivnak: Unfortunately, doing it in one line would not change the state of `u`. `u.Query()` returns a `Values` object, not a pointer to one, so it's a copy. Any changes made to that object are lost unless we do something like `u.RawQuery = q.Encode()`

registries/adapters/oauth/client.go
dymurray: I think he's referencing the `err check` PR https://github.com/automationbroker/bundle-lib/pull/95

registries/adapters/rhcc_adapter.go
mhrivnak: Seems reasonable. I think it's less likely here that we'll ever use more than one query parameter, whereas in the other case we could add more if dealing with different oauth implementations. But it wouldn't hurt to do it here, and that would be consistent.

registries/adapters/oauth/client.go
shawn-hurley: You're ~~Your~~ right, my apologies

registries/adapters/oauth/client.go
```diff
@@ -0,0 +1,233 @@
...
+// authResponse - holds the response data from an oauth2 token request
+type authResponse struct {
+	Token string `json:"access_token"`
+}
+
+// NewClient - creates and returns a *Client ready to use. If skipVerify is
+// true, it will skip verification of the remote TLS certificate.
+func NewClient(user, pass string, skipVerify bool, url *url.URL) *Client {
+	transport := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: skipVerify}}
+	if skipVerify == true {
```
jmrodri, line 43: I think `if skipVerify` is enough.

registries/adapters/oauth/client.go
jmrodri: @mhrivnak we have since closed PR #95 so do NOT do that. Leave it as is.

johnkim76 (approved):
Tested successfully using secret auth type

registries/adapters/oauth/client_test.go
jmrodri: +1 for the tests.

jmrodri (approved):
One really minor nit was the check for skipVerifyTLS == true instead of just if skipVerifyTLS. Not big enough to hold this PR hostage, so approving.

registries/adapters/oauth/client.go
mhrivnak: It looks like that PR is closed now, so does that mean there's nothing additional to do here?

shawn-hurley (approved):

PR #99, opened 2018-06-04
Adding ability to have the executor not create a namespace.

shawn-hurley:
* Adding Executor configuration struct.
* Changing the signature of new executor

### DIFF
--- a/bundle/bind.go
+++ b/bundle/bind.go
@@ -44,8 +44,12 @@ func (e *executor) Bind(
 
 	go func() {
 		e.actionStarted()
-		// Create namespace name that will be used to generate a name.
-		ns := fmt.Sprintf("%s-%.4s-", instance.Spec.FQName, bindAction)
+		ns := instance.Context.Namespace
+		// Determine if we should be using the context namespace from the executor config.
+		if !e.doNotCreateNS {
+			// Create namespace name that will be used to generate a name.
+			ns = fmt.Sprintf("%s-%.4s-", instance.Spec.FQName, bindAction)
+		}
 		// Create the podname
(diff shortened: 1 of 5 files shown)


mhrivnak (approved):
I hadn't noticed how we have nearly the same 80 lines of code repeated in 4 places. That's way beyond the scope of this PR, but would be interesting to chat about another time.

dymurray (approved):

bundle/bind.go
```diff
@@ -44,8 +44,12 @@ func (e *executor) Bind(
 
 	go func() {
 		e.actionStarted()
-		// Create namespace name that will be used to generate a name.
-		ns := fmt.Sprintf("%s-%.4s-", instance.Spec.FQName, bindAction)
+		ns := instance.Context.Namespace
+		// Determine if we should be using the context namespace from the executor config.
+		if !e.doNotCreateNS {
```
eriknelson: This is confusing phrasing with the double negative. I would do:

```
if e.doNotCreateNS {
ns := instance.Context.Namespace
else {
ns := createdNamespace
}
```

Less mental gymnastics and more direct.

eriknelson (changes requested):
Comment applies to all the conditional ns set statements, otherwise looks good.

bundle/bind.go
jmrodri: I prefer either: `createNS` OR `skipCreateNS`. Using skip allows you to use the Golang default of False for bools and to me it's easier to read then `doNot`

jmrodri (changes requested):
fix the double negative, otherwise +1

jmrodri (approved):

PR #102, opened 2018-06-07
Fix remove binding

jmrodri:
Pulled in the change @shawn-hurley had from release-0.1 and I added a test to it as well.


djzager (approved):
Thank you. Nice job adding a test.

PR #105, opened 2018-06-14
Use openshift client-go and api

shawn-hurley:
* Switch registry specific info to the adapter
* Move all of the openshift client calls to openshift-api and openshift-client-go
* adding test case for a single function in the openshift client.


clients/openshift.go
```diff
@@ -42,41 +40,9 @@ const (
...
-	// Items holds the image data
-	Items []image `json:"items"`
-}
-
-// FQImage is a struct to map FQNames to Imagestreams
-type FQImage struct {
-	Name        string
-	DecodedSpec []byte
-	Runtime     string
+	authClient    authv1.AuthorizationV1Interface
```
maleck13, line 43: awesome!

clients/openshift.go
```diff
@@ -265,12 +141,7 @@ func (o OpenshiftClient) GetClusterNetworkPlugin() (string, error) {
 
 // GetNetNamespace - Get Net Namespace.
 func (o OpenshiftClient) GetNetNamespace(nsName string) (*networkoapi.NetNamespace, error) {
-	netNamespace := &networkoapi.NetNamespace{}
-	err := o.networkClient.Get().Resource("netnamespaces").Name(nsName).Do().Into(netNamespace)
-	if err != nil {
-		return nil, err
-	}
-	return netNamespace, nil
+	return o.networkClient.NetNamespaces().Get(nsName, metav1.GetOptions{})
```
maleck13, line 144: Do you think the next logical step here is to expose the client interface directly as dependency and potentially remove the need for this wrapper at all

clients/openshift_test.go
```diff
@@ -0,0 +1,117 @@
...
+			namespace: "ns1",
+			shouldErr: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			c := authfake.Clientset{}
+			c.PrependReactor("create", "subjectrulesreviews", func(action clientgotesting.Action) (handled bool, ret runtime.Object, err error) {
+				logrus.Infof("%#v", action)
```
maleck13: do we want to keep this?

registries/adapters/local_openshift_adapter.go
```diff
@@ -74,71 +99,87 @@ func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*bundle.Spec,
 		return nil, err
 	}
 
-	fqImages, err := openshiftClient.ConvertRegistryImagesToSpecs(imageNames)
+	listImages, err := openshiftClient.ListRegistryImages()
 	if err != nil {
 		log.Errorf("Failed to load registry images")
```
maleck13, line 104: general comment about error handling in bundle-lib, but I think we should try to make use of the pkg/errors . This would mean that this kind of error handling would become:
```
return errors.Wrap(err,"Failed to load registery images")
```
This captures the context of the error and ensure what handles the error clearly sees the context. 
then the top level caller would decide what to do with the actual error. Log it , ignore, see the stacktrace ```log.Errorf("%+v", err) ``` the benefit would be reducing noise and handing control back to the consumer of the lib for this kind of thing.

I think I added an issue for this to the broker but will add one for bundle-lib too

registries/adapters/local_openshift_adapter.go
```diff
@@ -74,71 +99,87 @@ func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*bundle.Spec,
...
-			if ns == namespace {
-				log.Debugf("Image [%v] is in configured namespace [%v]. Adding to SpecList.", image.Name, ns)
-				specList = append(specList, spec)
+	for _, image := range listImages.Items {
+		n := strings.Split(image.DockerImageManifest, "@")[0]
+		for _, providedImage := range imageNames {
+			if providedImage == n {
+				spec, err := r.loadSpec(image)
+				if err != nil {
+					log.Errorf("Failed to load image spec")
```
maleck13, line 114: In cases like this , as this is a lib, do we want to use something like https://github.com/hashicorp/go-multierror or just implement something similar. Then hand back that set of errors to the consumer?

maleck13: If we did something like multi error, perhaps then we could group this logic into a  ```validateImage(image)error``` where the error here would be a multi error. Again just a general comment not really to do with the openshift client change

maleck13 (approved):
Looks good to me. Some general comment and thoughts but the change look fine

clients/openshift.go
shawn-hurley: @maleck13 I could see that, from my perspective the helper methods make the calling code "cleaner", especially the NetNamespaces ones. Images and Rules I could see differently but I am not rushing to remove the functions.

clients/openshift_test.go
shawn-hurley: Nope 😄 fixing that now thanks!

registries/adapters/local_openshift_adapter.go
shawn-hurley: @maleck13 If you think this is something that we should prioritize we should add it to the community meeting. We need to get a trello card then and start to try to get it on the backlog. 
I have not personally found the errors lacking in helping me track down errors, so I don't see the need to get this done ASAP. I would love to hear if others are having a different experience with bundle-lib so that we can make that better.

registries/adapters/local_openshift_adapter.go
shawn-hurley: @maleck13 I think if we can add this comment to the pkg/errors issue above and discuss on Tuesday? I think it is time to start thinking about the UX of the library.

maleck13: I will add it to the issue

registries/adapters/local_openshift_adapter.go
maleck13: No def not a big need to do it ASAP. more of a nice to have / UX consideration as you point out below

jmrodri (approved):
LGTM

dymurray (approved):

PR #109, opened 2018-06-15
create API V2 adapter

johnkim76:
modified partner_rhcc adapter
modified openshift adapter


johnkim76:
document [PR Link](https://github.com/openshift/ansible-service-broker/pull/992)

jmrodri:
This fails `make check` travis should totally have failed this.
```
$ make check
bundle/types.go:84:1: comment on exported type ServiceInstanceSchema should be of the form "ServiceInstanceSchema ..." (with optional leading article)
bundle/types.go:93:1: comment on exported type ServiceBindingSchema should be of the form "ServiceBindingSchema ..." (with optional leading article)
bundle/util.go:36:1: comment on exported function ConvertPlansToSchema should be of the form "ConvertPlansToSchema ..."
Found 3 lint suggestions; failing.
registries/adapters/apiv2_adapter.go:35:6: func NewApiV2Adapter should be NewAPIV2Adapter
registries/adapters/apiv2_adapter.go:43:6: type ApiV2Adapter should be APIV2Adapter
registries/adapters/apiv2_adapter.go:49:6: type ApiV2CatalogResponse should be APIV2CatalogResponse
registries/adapters/apiv2_adapter.go:89:38: method parameter catUrl should be catURL
registries/adapters/apiv2_adapter.go:94:26: error strings should not be capitalized or end with punctuation or a newline
registries/adapters/apiv2_adapter.go:158:23: method getNextImageUrl should be getNextImageURL
registries/adapters/apiv2_adapter_test.go:31:5: var ApiV2Images should be APIV2Images
registries/adapters/apiv2_adapter_test.go:33:7: const ApiV2ManifestResponse should be APIV2ManifestResponse
registries/adapters/apiv2_adapter_test.go:85:7: const ApiV2AuthResponse should be APIV2AuthResponse
Found 9 lint suggestions; failing.
go build -i -ldflags="-s -w" ./...
ok  	github.com/automationbroker/bundle-lib/bundle	0.038s	coverage: 39.3% of statements
ok  	github.com/automationbroker/bundle-lib/clients	(cached)	coverage: 24.4% of statements
?   	github.com/automationbroker/bundle-lib/crd	[no test files]
ok  	github.com/automationbroker/bundle-lib/registries	0.031s	coverage: 75.0% of statements
ok  	github.com/automationbroker/bundle-lib/registries/adapters	0.038s	coverage: 29.4% of statements
ok  	github.com/automationbroker/bundle-lib/registries/adapters/oauth	(cached)	coverage: 29.3% of statements
ok  	github.com/automationbroker/bundle-lib/runtime	(cached)	coverage: 29.4% of statements
?   	github.com/automationbroker/bundle-lib/runtime/mocks	[no test files]
```

jmrodri:
Looks like travis DID run make lint but didn't think it failed.

jmrodri:
I'm not sure I like this change. I like the consolidation but I don't like the loss of the actual registry adapter types. APIV2Adapter feels so generic and if we have a bug with a slight change for say partner registry then we'll have to fork it then. Basically I don't like the loss of the specific types.

I'd like to see something like

```golang
type OpenShiftAdapter struct {
    APIV2Adapter
}
```

I believe this will allow OpenShiftAdapter to still implement the `Registry` adapter but retain the type. I'll defer to @shawn-hurley for the correct syntax on doing what I'm suggesting.

registries/adapters/apiv2_adapter.go
```diff
@@ -0,0 +1,220 @@
...
+	"io/ioutil"
+	"strings"
+
+	"github.com/automationbroker/bundle-lib/bundle"
+	"github.com/automationbroker/bundle-lib/registries/adapters/oauth"
+	log "github.com/sirupsen/logrus"
+)
+
+const apiV2ManifestURL = "%v/v2/%v/manifests/%v"
+const apiV2CatalogURL = "%v/v2/_catalog"
```
jmrodri: Do we really need to call it the `apiV2CatalogURL`? isn't it just the catalog url? It is part of the APIV2Adapter presumably we would  never have an apiV3CatalogURL be part of the APIV2Adapter.

I'd keep these just as catalogURL and manifestURL.

registries/adapters/apiv2_adapter.go
```diff
@@ -0,0 +1,220 @@
...
+
+// NewApiV2Adapter - creates and returns a *ApiV2Adapter ready to use.
+func NewApiV2Adapter(config Configuration) *ApiV2Adapter {
```
jmrodri: ApiV2 won't pass linter. It will need to be APIV2Adapter, which I don't really care for but I don't have another suggestion at the time.

registries/adapters/apiv2_adapter_test.go
```diff
@@ -24,12 +24,13 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/automationbroker/bundle-lib/registries/adapters/oauth"
 	ft "github.com/stretchr/testify/assert"
 )
 
-var Images = []string{"satoshi/nakamoto", "foo/bar", "paul/atreides"}
+var ApiV2Images = []string{"satoshi/nakamoto", "foo/bar", "paul/atreides"}
```
jmrodri: I would leave this as `Images`

registries/adapters/apiv2_adapter_test.go
```diff
@@ -152,16 +159,15 @@ func TestOpenShiftFetchSpecs(t *testing.T) {
 	if err != nil {
 		t.Fatal("Error: ", err)
 	}
-	ocpa := OpenShiftAdapter{
+	apiv2a := ApiV2Adapter{
```
jmrodri: v2adapter is probably what I would've just call this variable or just adapter.

registries/adapters/openshift_adapter.go
```diff
@@ -1,162 +0,0 @@
...
-	"strings"
-
-	"github.com/automationbroker/bundle-lib/bundle"
-	log "github.com/sirupsen/logrus"
-)
-
-const openShiftManifestURL = "%v/v2/%v/manifests/%v"
-
-// OpenShiftAdapter - Docker Hub Adapter
-type OpenShiftAdapter struct {
```
jmrodri, line 33: -10000 I don't like the loss of the `OpenShiftAdatper` type.

registries/adapters/partner_rhcc_adapter.go
```diff
@@ -1,193 +0,0 @@
...
-	"github.com/automationbroker/bundle-lib/bundle"
-	log "github.com/sirupsen/logrus"
-)
-
-const partnerName = "partner_rhcc"
-const partnerManifestURL = "%v/v2/%v/manifests/%v"
-const partnerCatalogURL = "%v/v2/_catalog"
-
-// PartnerRhccAdapter - Partner RHCC Adapter
-type PartnerRhccAdapter struct {
```
jmrodri, line 36: -10000 I don't like the loss of the PartnerRhccAdapter type.

jmrodri (changes requested):
I like the refactoring of code to reduce duplication but I don't like the loss of the actual types.

registries/adapters/apiv2_adapter.go
```diff
@@ -0,0 +1,220 @@
...
+
+// discoverImages - Get Imagenames from the /v2/_catalog URL
+func (r ApiV2Adapter) discoverImages(catUrl string) ([]string, error) {
+	log.Debugf("%s - discoverImages", r.Config.AdapterName)
+
+	// Initial URL
+	if len(catUrl) == 0 {
+		return nil, errors.New("URL is empty!")
+	}
+	url := catUrl
```
jmrodri: This seems like a useless assignment. Remove this line. Then change line 89 to read:
`discoverImages(url string) ...`

registries/adapters/apiv2_adapter.go
shawn-hurley: But v3 adapter would most likely be in the same package and therefore would need the v2 prefix to identify which one inside the package.

With the above said I don't see a v3 coming anytime soon so if you feel strongly then I am good with removing the version prefix.

registries/adapters/apiv2_adapter.go
```diff
@@ -0,0 +1,220 @@
...
+		return nil, "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		log.Errorf("Failed to fetch catalog response from '%s'. Expected a 200 status and got: %v", url, resp.Status)
+		return nil, "", errors.New(resp.Status)
+	}
+
+	imageResp, err := ioutil.ReadAll(resp.Body)
```
shawn-hurley: This is a nit because this is perfectly fine.

There are helper methods to do this
```go
imageList := ApiV2CatalogResponse{}
err = json.NewDecoder(resp.Body).Decode(&imageList)
```

registries/adapters/apiv2_adapter.go
```diff
@@ -0,0 +1,220 @@
...
+	return r.Config.URL.Host
+}
+
+// GetImageNames - retrieve the images
+func (r ApiV2Adapter) GetImageNames() ([]string, error) {
+	log.Debugf("%s - GetImageNames", r.Config.AdapterName)
+	log.Debugf("BundleSpecLabel: %s", BundleSpecLabel)
+
+	// Authorization
+	err := r.client.Getv2()
```
shawn-hurley: We need to initialize this before any call to the registry. I am wondering if this should maybe be don in the new function. This would make it so that a user does not have to call GetImagesNames before anything else?

shawn-hurley (changes requested):

shawn-hurley:
I agree with @jmrodri about the openshift adapter type that just embeds the `APIAdapter`. 

I think this is a really good change is very readable when it comes to the logic of the adapter, thank you!

jmrodri:
@shawnhurley your point about v3 makes sense, I didn't think of that. keep it as v2 then.

johnkim76:
thanks @jmrodri and @shawn-hurley for input, I will make an update with all things addressed

shawn-hurley (approved):
Everything looks good from me: Just this one NIT: https://github.com/automationbroker/bundle-lib/pull/109/files#r196067853

johnkim76:
@shawn-hurley sorry missed that last one... fixed.

jmrodri (approved):

registries/adapters/apiv2_adapter.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/automationbroker/bundle-lib/registries/adapters/oauth"
+	log "github.com/sirupsen/logrus"
+)
+
+const apiV2ManifestURL = "%v/v2/%v/manifests/%v"
+const apiV2CatalogURL = "%v/v2/_catalog"
+
+// OpenShiftAdapter - OpenShift Adapter
+type OpenShiftAdapter struct {
+	*apiV2Adapter
```
eriknelson: There a specific reason these are getting embedded as pointers? It seems like these should't be getting passed around and mutated after their construction.

registries/adapters/apiv2_adapter.go
```diff
@@ -0,0 +1,245 @@
...
+
+// GetImageNames - retrieve the images
+func (r apiV2Adapter) GetImageNames() ([]string, error) {
+	log.Debugf("%s - GetImageNames", r.config.AdapterName)
+	log.Debugf("BundleSpecLabel: %s", BundleSpecLabel)
+
+	// check if we're configured for specific images
+	if r.config.Images != nil {
+		log.Debugf("Configured to use images: %v", r.config.Images)
+		return r.config.Images, nil
```
dymurray: This is fine for now but I think we wanted to also be able to include configured images AND try discovery. Thoughts?

shawn-hurley: Should this be an issue to track this idea?

johnkim76: @shawn-hurley no i don't think that'll be necessary.  as discussed with @dymurray I will do the following
- Add custom Images if list is specified 
- do discovery, and append to the image list
- If there are no images after the above two steps, return error
What do y'all think?

registries/adapters/apiv2_adapter.go
johnkim76: @eriknelson no, I was just following the recent example that @mhrivnak added to the [rhcc_adaptor](https://github.com/automationbroker/bundle-lib/blob/master/registries/adapters/rhcc_adapter.go#L31)
I don't have a real good reason for it, I will remove it

johnkim76:
@dymurray @eriknelson I've updated as suggested

dymurray (approved):

jmrodri:
I reapprove the changes.

registries/adapters/apiv2_adapter.go
```diff
@@ -0,0 +1,249 @@
...
+const apiV2ManifestURL = "%v/v2/%v/manifests/%v"
+const apiV2CatalogURL = "%v/v2/_catalog"
+
+// OpenShiftAdapter - OpenShift Adapter
+type OpenShiftAdapter struct {
+	apiV2Adapter
+}
+
+// PartnerRhccAdapter - Partner RHCC Adapter
+type PartnerRhccAdapter struct {
```
mhrivnak: What's the thinking behind keeping the multiple types? I figured there would be something like `APIv2Adapter` that gets exported and meets the interface obligations of an adapter. The types are different in name only; all behavior is identical. So from a software perspective, what's the value in having separate types?

jmrodri: Well mostly because I whined about it. I preferred having the actual types vs just using a generic type. Probably for preoptimization, but I suspect at some point you'll need to fix something for PartnerRhccAdapter that isn't supposed to be fixed in DockerHub. But to me from a maintenance point of view. If a bug comes in stating they're having trouble connecting to PartnerRhcc, I know EXACTLY where to look. Vs looking for the config to find the magic string, then finding out it's APIV2, then find APIV2. Also, it's almost a form of inheritance (poor choice of word for a language that doesn't have it, but still kinda is), PartnerRhcc AND DockerHub *are* APIV2 Adapters.

dymurray: From a CLI standpoint I would prefer for someone to just be able to instantiate a `v2` adapter with a URL and whitelist and expect it to work. It's a pain having to validate the types and instantiate different defaults depending on the type. I am in favor of leaving the types so that backwards compatibility works but I would prefer to instantiate the same type of adapter each time.

johnkim76: @jmrodri, would you object to exporting `apiV2Adapter` and add a `v2adapter` type to `registry.go`?

jmrodri: @johnkim76 whatever, I've said my peace, expose APIV2. @dymurray as a user I know I'm trying to connect to openshift registry or partner registry. I don't really *know* they're APIV2. But we've gone on long enough, you guys have spent more time in registry stuff.

johnkim76:
@jmrodri @dymurray @mhrivnak I've updated the PR to export `APIV2Adapter` and added a `apiv2` adapter type.

mhrivnak (approved):
Nicely done.

johnkim76:
since we decided that a user can use the `apiv2` directly.. here's the documentation for it
https://github.com/openshift/ansible-service-broker/pull/1003

dymurray (approved):
Reapprove

PR #113, opened 2018-06-21
Expose route interface from OpenShift client

dymurray:
This gives us the ability to get a RouteList object given a namespace. This is useful for `sbcli` since we need to get the broker's route.

### DIFF
--- a/clients/openshift.go
+++ b/clients/openshift.go
@@ -23,7 +23,6 @@ import (
 	authoapi "github.com/openshift/api/authorization/v1"
 	"github.com/openshift/api/image/v1"
 	networkoapi "github.com/openshift/api/network/v1"
-	routeoapi "github.com/openshift/api/route/v1"
 	authv1 "github.com/openshift/client-go/authorization/clientset/versioned/typed/authorization/v1"
 	imagev1 "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	networkv1 "github.com/openshift/client-go/network/clientset/versioned/typed/network/v1"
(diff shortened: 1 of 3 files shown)


djwhatle (approved):
LGTM

clients/openshift.go
```diff
@@ -23,9 +23,11 @@ import (
 	authoapi "github.com/openshift/api/authorization/v1"
 	"github.com/openshift/api/image/v1"
 	networkoapi "github.com/openshift/api/network/v1"
+	routeoapi "github.com/openshift/api/route/v1"
 	authv1 "github.com/openshift/client-go/authorization/clientset/versioned/typed/authorization/v1"
 	imagev1 "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	networkv1 "github.com/openshift/client-go/network/clientset/versioned/typed/network/v1"
+	routev1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
```
jmrodri: So the two imports are intentional? Just make sure since I can see how goimports would confuse the two.

jmrodri (approved):
ACK assuming the imports were intentional.

clients/openshift.go
dymurray: Yeah they are intentional. Was following the approach for the `network`, `image`, and `auth` resources.

dymurray:
Some thoughts although I'd like to merge this since I need it for `sbcli`...
I'm curious if you guys think our `openshift` client package will get too bloated if we continue down this approach. It felt weird to add a specific resource getter (which can be accomplished by just importing `openshift/client-go` code into `scbli`) here but I wanted `sbcli` to completely consume bundle-libs `openshift`client package so I felt this belonged here. Ideally we write some helper functions that allow us to scale this so it can be easily consumed by our other projects but isn't just adding new very specific needed functions.

shawn-hurley:
@dymurray could we add a single test? I know it seems trivial but we need to get the code coverage up.

shawn-hurley:
@dymurray I think it could be as easy as this, and then we only keep helper functions where they make sense. I.E I think the net namespaces helper functions are helpful because you can call a named function (add, remove) and we handle the details of the annotation. 

```go
func (o OpenshiftClient)  Route() routev1.RouteV1Interface {
	return o.routeClient
}
```

clients/openshift.go
```diff
@@ -175,3 +181,8 @@ func (o OpenshiftClient) IsolateNamespacesNetworks(netns *networkoapi.NetNamespa
 	}
 	return result, nil
 }
+
+// Route - Returns a V1Route Interface
+func (o OpenshiftClient) Route() *routev1.RouteV1Interface {
```
shawn-hurley: 👍

shawn-hurley (approved):
Thanks for the test!

PR #115, opened 2018-06-28
Update local_openshift adapter to use imagestreams as opposed to images

dymurray:
Needed for `sbcli` since users don't always have access to do `oc get images`

### DIFF
--- a/registries/adapters/local_openshift_adapter.go
+++ b/registries/adapters/local_openshift_adapter.go
@@ -73,7 +73,7 @@ func (r LocalOpenShiftAdapter) GetImageNames() ([]string, error) {
 
 	imageClient := openshiftClient.Image()
 	if r.Config.Namespaces == nil {
-		log.Debug("Didn't find any namespace in broker configuration, assuming `openshift`.")
+		log.Debug("Didn't find any namespace in configuration, assuming `openshift`.")
 		r.Config.Namespaces = append(r.Config.Namespaces, "openshift")
 	}
 	imageList := []string{}
@@ -102,14 +102,19 @@ func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*bundle.Spec, 
 	}
 	imageClient := openshiftClient.Image()
 
+	if r.Config.Tag == "" {
+		log.Debug("No tag specified in config, assuming `latest`")
+		r.Config.Tag = "latest"
+	}
+
 	for _, image := range imageNames {
 		fullName := strings.Split(image, "/")
+		if len(fullName) < 2 {
+			log.Errorf("Image name [%v] not in expected format, skipping.", image)
+			continue
+		}
 		ns := fullName[0]
 		iName := fullName[1]
-		if r.Config.Tag == "" {
-			log.Debug("No tag specified in config, assuming `latest`")
-			r.Config.Tag = "latest"
-		}
 		imTag, err := imageClient.ImageStreamTags(ns).Get(fmt.Sprintf("%v:%v", iName, r.Config.Tag), meta_v1.GetOptions{})
 		if err != nil {
 			log.Errorf("Failed to get image for imagestream [%v]: %v", image, err)
@@ -118,6 +123,7 @@ func (r LocalOpenShiftAdapter) FetchSpecs(imageNames []string) ([]*bundle.Spec, 
 		spec, err := r.loadSpec(imTag.Image)
 		if err != nil {
 			log.Errorf("Failed to load spec for [%v]: %v", image, err)
+			continue
 		}
 		specList = append(specList, spec)
 	}
--- a/registries/adapters/local_openshift_adapter.go
+++ b/registries/adapters/local_openshift_adapter.go
@@ -19,8 +19,10 @@ package adapters
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"strings"
 
+	b64 "encoding/base64"
 	"github.com/automationbroker/bundle-lib/bundle"
 	"github.com/automationbroker/bundle-lib/clients"
 	v1image "github.com/openshift/api/image/v1"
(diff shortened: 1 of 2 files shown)


registries/adapters/local_openshift_adapter.go
```diff
@@ -170,7 +140,12 @@ func (r LocalOpenShiftAdapter) loadSpec(image v1image.Image) (*bundle.Spec, erro
 	}
 	spec := &bundle.Spec{}
 
-	err = yaml.Unmarshal([]byte(i.ContainerConfig.Labels.Spec), spec)
+	decodedSpecYaml, err := b64.StdEncoding.DecodeString(i.ContainerConfig.Labels.Spec)
```
djwhatle, line 149: Interesting that we have to do a decode here where we previously didn't.

registries/adapters/local_openshift_adapter.go
```diff
@@ -69,87 +71,55 @@ func (r LocalOpenShiftAdapter) GetImageNames() ([]string, error) {
...
-						specList = append(specList, spec)
-					}
-				}
-			}
+	imageClient := openshiftClient.Image()
+
+	for _, image := range imageNames {
+		fullName := strings.Split(image, "/")
+		ns := fullName[0]
+		iName := fullName[1]
```
djwhatle: Array length check needed here? Or do we have high confidence that the data we're ingesting is formatted this way

registries/adapters/local_openshift_adapter.go
```diff
@@ -69,87 +71,55 @@ func (r LocalOpenShiftAdapter) GetImageNames() ([]string, error) {
...
-	for _, image := range images.Items {
-		imageList = append(imageList, strings.Split(image.DockerImageManifest, "@")[0])
+	for _, ns := range r.Config.Namespaces {
+		is, err := imageClient.ImageStreams(ns).List(meta_v1.ListOptions{})
+		if err != nil {
+			log.Errorf("Failed to get list of imagestreams for namespace [%v]: %v", ns, err)
+			continue
+		}
+		for _, i := range is.Items {
+			imageList = append(imageList, fmt.Sprintf("%v/%v", ns, i.Name))
```
djwhatle, line 87: Curious about implications (are there any?) of using `%v` in places where we expect a string. Do you get anything valuable out of using `%s` instead? I don't know the answer.

registries/adapters/local_openshift_adapter.go
```diff
@@ -69,87 +71,55 @@ func (r LocalOpenShiftAdapter) GetImageNames() ([]string, error) {
 		return nil, err
 	}
 
-	images, err := openshiftClient.ListRegistryImages()
-	if err != nil {
-		log.Errorf("Failed to load registry images")
-		return nil, err
+	imageClient := openshiftClient.Image()
+	if r.Config.Namespaces == nil {
+		log.Debug("Didn't find any namespace in broker configuration, assuming `openshift`.")
```
djwhatle: Is it appropriate to use the term "broker configuration" here, since bundle-lib isn't only used by the broker?

djwhatle (commented):
Made some comments, don't necessarily see anything needing changes. Want to test this out with sbcli. Do you mind posting some instructions on how to do so? (what sbcli commands to run)

djwhatle:
I was able to verify that this works with `sbcli` for viewing specs created from OpenShift builds in unprivileged namespaces, and also verified that provision works as expected.

djwhatle (approved):
Worked for sbcli viewing and provisioning of APBs built in local namespace.

registries/adapters/local_openshift_adapter.go
jmrodri: I would just remove broker from that line.

registries/adapters/local_openshift_adapter.go
jmrodri: @djwhatle %v is almost always okay to use regardless of the type.

```
%v	the value in a default format
	when printing structs, the plus flag (%+v) adds field names
```

registries/adapters/local_openshift_adapter.go
jmrodri: The godoc states:

> func Split(s, sep string) []string
>
> Split slices s into all substrings separated by sep and returns a slice of the substrings between those separators.
>
> If s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s.
>
> If sep is empty, Split splits after each UTF-8 sequence. If both s and sep are empty, Split returns an empty slice.
>
> It is equivalent to SplitN with a count of -1. 

Just tried it, if the string does not have a "/" in it, you won't get a second element in the array it will cause the broker to panice.

registries/adapters/local_openshift_adapter.go
```diff
@@ -69,87 +71,55 @@ func (r LocalOpenShiftAdapter) GetImageNames() ([]string, error) {
...
+	imageClient := openshiftClient.Image()
+
+	for _, image := range imageNames {
+		fullName := strings.Split(image, "/")
+		ns := fullName[0]
+		iName := fullName[1]
+		if r.Config.Tag == "" {
+			log.Debug("No tag specified in config, assuming `latest`")
+			r.Config.Tag = "latest"
+		}
```
jmrodri: Feels weird to do this Config.Tag check every loop. I think this should be done outside the loop.

registries/adapters/local_openshift_adapter.go
```diff
@@ -69,87 +71,55 @@ func (r LocalOpenShiftAdapter) GetImageNames() ([]string, error) {
...
+		imTag, err := imageClient.ImageStreamTags(ns).Get(fmt.Sprintf("%v:%v", iName, r.Config.Tag), meta_v1.GetOptions{})
+		if err != nil {
+			log.Errorf("Failed to get image for imagestream [%v]: %v", image, err)
+			continue
 		}
+		spec, err := r.loadSpec(imTag.Image)
+		if err != nil {
+			log.Errorf("Failed to load spec for [%v]: %v", image, err)
+		}
+		specList = append(specList, spec)
```
jmrodri: This will still add the spec even if the loadSpec fails. This might be adding an invalid spec to the list. Is that what we want to do?

registries/adapters/local_openshift_adapter.go
```diff
@@ -180,6 +155,7 @@ func (r LocalOpenShiftAdapter) loadSpec(image v1image.Image) (*bundle.Spec, erro
 		log.Errorf("Failed to parse image runtime version")
 		return nil, errRuntimeNotFound
 	}
+	spec.Image = strings.Split(image.DockerImageReference, "@")[0]
```
jmrodri, line 164: this is okay, Split will always return at least one element.

jmrodri (changes requested):
Some questions and comments.

registries/adapters/local_openshift_adapter.go
dymurray: Nope, thanks for catching this!

djwhatle: I wonder if this is related to the issue I saw here: https://github.com/automationbroker/sbcli/issues/68

jmrodri (approved):

PR #121, opened 2018-07-05
Change logs from INFO to DEBUG to clean up sbcli output

djwhatle:
Changed some log statements to use `DEBUG` log level instead of `INFO` in order to clean up sbcli output. For the creation of clientsets, I believe we will still get an error if all attempts fail.

`sbcli provision` __without this PR__
```
WARN Found multiple bundles matching name [v2v]. Specify a registry with --registry. 
[~/code/go/src/github.com/automationbroker/sbcli] derek $ >> sbcli bundle provision v2v --registry docker2
Found bundle [v2v] in registry [docker2]
Plan: default
Enter value for parameter [admin_user], default: [<nil>]:  foo
[...]
WARN Failed to create a InternalClientSet: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined. 
INFO OpenShift version: v3.11.0-alpha.0+60cb624-216-dirty 
WARN Failed to create a InternalClientSet: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined. 
INFO unable to retrieve the network plugin, defaulting to not joining networks - clusternetworks.network.openshift.io "default" not found 
INFO Creating RoleBinding bundle-06401245-7c42-479e-892c-4b1dd9aafe4f 
INFO Successfully created apb sandbox: [ bundle-06401245-7c42-479e-892c-4b1dd9aafe4f ], with edit permissions in namespace foobar 
INFO Running post create sandbox functions if defined. 
Successfully created pod [bundle-06401245-7c42-479e-892c-4b1dd9aafe4f] to provision [v2v] in namespace [foobar]
[~/code/go/src/github.com/automationbroker/sbcli] derek $ >> 
```

`sbcli provision` __with this PR__ :

```
[~/code/go/src/github.com/automationbroker/bundle-lib] derek $ >> sbcli bundle provision v2v -n foobar --registry docker
Found bundle [v2v] in registry [docker]
Plan: default
Enter value for parameter [admin_user], default: [<nil>]: foo
[...]
INFO Creating RoleBinding bundle-ba3947ed-7bbb-4161-a95e-84cd2b96fa48 
INFO Successfully created apb sandbox: [ bundle-ba3947ed-7bbb-4161-a95e-84cd2b96fa48 ], with edit permissions in namespace [ foobar ] 
Successfully created pod [bundle-ba3947ed-7bbb-4161-a95e-84cd2b96fa48] to provision [v2v] in namespace [foobar]
```

Notice how we don't get any more "Failed to create InternalClientSet" errors.

### DIFF
--- a/clients/crd.go
+++ b/clients/crd.go
@@ -52,8 +52,6 @@ func newCRDClient() (*CRD, error) {
 	// library
 	clientConfig, err := rest.InClusterConfig()
 	if err != nil {
-		log.Debugf("Failed to create a InternalClientSet: %v.", err)
-
 		log.Debug("Checking for a local Cluster Config")
 		clientConfig, err = createClientConfigFromFile(homedir.HomeDir() + "/.kube/config")
 		if err != nil {
--- a/clients/crd.go
+++ b/clients/crd.go
@@ -52,7 +52,7 @@ func newCRDClient() (*CRD, error) {
 	// library
 	clientConfig, err := rest.InClusterConfig()
 	if err != nil {
-		log.Warningf("Failed to create a InternalClientSet: %v.", err)
+		log.Debugf("Failed to create a InternalClientSet: %v.", err)
 
 		log.Debug("Checking for a local Cluster Config")
 		clientConfig, err = createClientConfigFromFile(homedir.HomeDir() + "/.kube/config")
--- a/clients/kubernetes.go
+++ b/clients/kubernetes.go
@@ -209,7 +209,7 @@ func newKubernetes() (*KubernetesClient, error) {
 	// library
 	clientConfig, err := rest.InClusterConfig()
 	if err != nil {
-		log.Warningf("Failed to create a InternalClientSet: %v.", err)
+		log.Debugf("Failed to create a InternalClientSet: %v.", err)
 
 		log.Debug("Checking for a local Cluster Config")
 		clientConfig, err = createClientConfigFromFile(homedir.HomeDir() + "/.kube/config")
--- a/clients/kubernetes.go
+++ b/clients/kubernetes.go
@@ -209,8 +209,6 @@ func newKubernetes() (*KubernetesClient, error) {
 	// library
 	clientConfig, err := rest.InClusterConfig()
 	if err != nil {
-		log.Debugf("Failed to create a InternalClientSet: %v.", err)
-
 		log.Debug("Checking for a local Cluster Config")
 		clientConfig, err = createClientConfigFromFile(homedir.HomeDir() + "/.kube/config")
 		if err != nil {
--- a/clients/openshift.go
+++ b/clients/openshift.go
@@ -74,8 +74,6 @@ func newOpenshift() (*OpenshiftClient, error) {
 	// library
 	clientConfig, err := rest.InClusterConfig()
 	if err != nil {
-		log.Debugf("Failed to create a InternalClientSet: %v.", err)
-
 		log.Debug("Checking for a local Cluster Config")
 		clientConfig, err = createClientConfigFromFile(homedir.HomeDir() + "/.kube/config")
 		if err != nil {
--- a/clients/openshift.go
+++ b/clients/openshift.go
@@ -74,7 +74,7 @@ func newOpenshift() (*OpenshiftClient, error) {
 	// library
 	clientConfig, err := rest.InClusterConfig()
 	if err != nil {
-		log.Warningf("Failed to create a InternalClientSet: %v.", err)
+		log.Debugf("Failed to create a InternalClientSet: %v.", err)
 
 		log.Debug("Checking for a local Cluster Config")
 		clientConfig, err = createClientConfigFromFile(homedir.HomeDir() + "/.kube/config")
--- a/runtime/openshift.go
+++ b/runtime/openshift.go
@@ -45,7 +45,7 @@ func (o openshift) shouldJoinNetworks() (bool, PostSandboxCreate, PostSandboxDes
 	if err != nil {
 		// The plugins could not be defined. ex: when using oc cluster up
 		// or a pure k8s cluster. Therefore making this a notice.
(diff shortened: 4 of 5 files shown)


dymurray (approved):

clients/kubernetes.go
```diff
@@ -209,7 +209,7 @@ func newKubernetes() (*KubernetesClient, error) {
 	// library
 	clientConfig, err := rest.InClusterConfig()
 	if err != nil {
-		log.Warningf("Failed to create a InternalClientSet: %v.", err)
+		log.Debugf("Failed to create a InternalClientSet: %v.", err)
```
djwhatle: In case anyone is concerned with this warning getting buried and the user not getting needed info to diagnose an issue, line 217 should still print an error if all attempts to create a client config fail.

clients/openshift.go
```diff
@@ -74,7 +74,7 @@ func newOpenshift() (*OpenshiftClient, error) {
 	// library
 	clientConfig, err := rest.InClusterConfig()
 	if err != nil {
-		log.Warningf("Failed to create a InternalClientSet: %v.", err)
+		log.Debugf("Failed to create a InternalClientSet: %v.", err)
```
djwhatle: Same deal here, line 83 should handle printing a user-visible error message, so my take is that this log statement being a "Warning" isn't critical

runtime/openshift.go
```diff
@@ -45,7 +45,7 @@ func (o openshift) shouldJoinNetworks() (bool, PostSandboxCreate, PostSandboxDes
 	if err != nil {
 		// The plugins could not be defined. ex: when using oc cluster up
 		// or a pure k8s cluster. Therefore making this a notice.
-		log.Infof("unable to retrieve the network plugin, defaulting to not joining networks - %v", err)
+		log.Debugf("unable to retrieve the network plugin, defaulting to not joining networks - %v", err)
```
djwhatle, line 48: This log statement is stated to be non-critical in the comment above, so figured it was ok to set it to "Debug"

runtime/runtime.go
```diff
@@ -117,7 +117,7 @@ func NewRuntime(config Configuration) {
 			log.Error(err.Error())
 			panic(err.Error())
 		}
-		log.Infof("OpenShift version: %v", kubeServerInfo)
+		log.Debugf("OpenShift version: %v", kubeServerInfo)
```
jmrodri: This is useful in the broker logs at info level. Now we'll have to turn on debug to see this useful information. Feels like we're making these changes when maybe bundle-lib should be configured to use a different loglevel.

runtime/runtime.go
```diff
@@ -348,8 +348,8 @@ func (p provider) CreateSandbox(podName string,
 	}
 	log.Debugf("Successfully created network policy for pod: %v to grant network access to ns: %v", podName, targets[0])
 
-	log.Infof("Successfully created apb sandbox: [ %s ], with %s permissions in namespace %s", podName, apbRole, namespace)
-	log.Info("Running post create sandbox functions if defined.")
+	log.Infof("Successfully created apb sandbox: [ %s ], with %s permissions in namespace [ %s ]", podName, apbRole, namespace)
+	log.Debug("Running post create sandbox functions if defined.")
```
jmrodri, line 352: this is okay.

runtime/openshift.go
jmrodri: I really think we should be checking the error before eating the log.  But then again we're not doing anything with the error `return false, nil, nil`. Feels like we're really ignore a real error.

clients/openshift.go
jmrodri: I would advocate to just remove these message altogether. Because like you said line 83 will print the real error. I don't like seeing a statement "Failed to create " as debug.

clients/kubernetes.go
jmrodri: remove log altogether.

clients/crd.go
```diff
@@ -52,7 +52,7 @@ func newCRDClient() (*CRD, error) {
 	// library
 	clientConfig, err := rest.InClusterConfig()
 	if err != nil {
-		log.Warningf("Failed to create a InternalClientSet: %v.", err)
+		log.Debugf("Failed to create a InternalClientSet: %v.", err)
```
jmrodri: remove log altogether.

jmrodri (changes requested):
Remove the InternalClientSet logs altogether. Then some of the others where we log the error as debug and then return false, nil, nil, it feels like we should really be looking at the error. If the err is ignoreable we return false, nil, nil and no log. If it really is an error, we should do something with it.

runtime/runtime.go
djwhatle: I'm actually OK with keeping this one in because its a pretty clean log, will change it back to Info. 

What do you mean bundle-lib should be configured to use a different loglevel? Do you mean `apb` should be configured to use a different loglevel?

runtime/openshift.go
djwhatle: I'm not sure what errors might result here / how we would handle them. I can file an issue if that would make sense.

eriknelson: I think this should be of a higher log level than a debug. Maybe Info, probably not Warning. How often do we expect this to occur?

djwhatle: It happens during every `apb` provision/deprovision, at least with the 3.9 cluster I'm using.

dymurray: @eriknelson In the CLI tool since we are executing against an `oc cluster up` environment frequently it shows up 100% of the time on INFO.

shawn-hurley (approved):

shawn-hurley (approved):

dymurray:
Opened https://github.com/automationbroker/apb/issues/82

jmrodri (approved):

PR #124, opened 2018-07-06
Add galaxy adapter

djzager:
Convert @jmontleon `galaxy-adapter` branch of ansible-service-broker into a PR for bundle-lib.

### DIFF
--- a/registries/adapters/galaxy_adapter.go
+++ b/registries/adapters/galaxy_adapter.go
@@ -85,7 +85,7 @@ func (r GalaxyAdapter) RegistryName() string {
 }
 
 // GetImageNames - retrieve the images
-func (r GalaxyAdapter) GetImageNames() ([]string, error) {
+func (r *GalaxyAdapter) GetImageNames() ([]string, error) {
 	log.Debug("GalaxyAdapter::GetImages")
 	log.Debugf("BundleSpecLabel: %s", BundleSpecLabel)
 	log.Debug("Loading role list with tag: [apb]")
@@ -144,6 +144,13 @@ func (r GalaxyAdapter) GetImageNames() ([]string, error) {
 
 // FetchSpecs - retrieve the spec for the image names.
 func (r GalaxyAdapter) FetchSpecs(imageNames []string) ([]*bundle.Spec, error) {
+
+	// default galaxy url
+	if r.Config.URL.Host == "" {
+		log.Debugf("Using default galaxy url: %v", defaultURL)
+		r.Config.URL, _ = url.Parse(defaultURL)
+	}
+
 	specs := []*bundle.Spec{}
 	for _, imageName := range imageNames {
 		spec, err := r.loadSpec(imageName)
--- a/registries/adapters/galaxy_adapter.go
+++ b/registries/adapters/galaxy_adapter.go
@@ -22,19 +22,20 @@ import (
 	"fmt"
 	"github.com/automationbroker/bundle-lib/bundle"
 	log "github.com/sirupsen/logrus"
-	//yaml "gopkg.in/yaml.v1"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 )
 
 const (
 	defaultRunner     = "ansibleplaybookbundle/apb-base:latest"
+	defaultURL        = "https://galaxy.ansible.com"
 	galaxyName        = "galaxy"
-	galaxySearchURL   = "https://galaxy-qa.ansible.com/api/v1/content/?content_type__name=apb"
-	galaxyNSSearchURL = "https://galaxy-qa.ansible.com/api/v1/content/?content_type__name=apb&namespace__name=%v"
-	galaxyRoleURL     = "https://galaxy-qa.ansible.com/api/v1/content/%v/"
-	galaxyAPIURL      = "https://galaxy-qa.ansible.com/api/v1"
+	galaxySearchURL   = "%v/api/v1/content/?content_type__name=apb"
+	galaxyNSSearchURL = "%v/api/v1/content/?content_type__name=apb&namespace__name=%v"
+	galaxyRoleURL     = "%v/api/v1/content/%v/"
+	galaxyAPIURL      = "%v/api/v1%v"
 )
 
 // GalaxyAdapter - Galaxy Adapter
@@ -89,18 +90,29 @@ func (r GalaxyAdapter) GetImageNames() ([]string, error) {
 	log.Debug("BundleSpecLabel: %s", BundleSpecLabel)
 	log.Debug("Loading role list with tag: [apb]")
 
+	// default galaxy url
+	if r.Config.URL == nil {
+		log.Debugf("Using default galaxy url: %v", defaultURL)
+		r.Config.URL, _ = url.Parse(defaultURL)
+	}
+
+	// generate search url
+	searchURL := fmt.Sprintf(galaxySearchURL, r.Config.URL.String())
+	if len(r.Config.Org) != 0 {
+		log.Debugf("Using namespaced galaxy search")
+		searchURL = fmt.Sprintf(galaxyNSSearchURL, r.Config.URL.String(), r.Config.Org)
+	}
+
 	channel := make(chan string)
 	ctx, cancelFunc := context.WithCancel(context.Background())
 	defer cancelFunc()
 
 	var imageResp *GalaxySearchResponse
 	var err error
+
 	// Intial call to getNextImages this will fan out to retrieve all the values.
-	if len(r.Config.Org) == 0 {
-		imageResp, err = r.getNextImages(ctx, galaxySearchURL, channel, cancelFunc)
-	} else {
-		imageResp, err = r.getNextImages(ctx, fmt.Sprintf(galaxyNSSearchURL, r.Config.Org), channel, cancelFunc)
-	}
+	imageResp, err = r.getNextImages(ctx, searchURL, channel, cancelFunc)
+
 	// if there was an issue with the first call, return the error
 	if err != nil {
 		return nil, err
@@ -180,8 +192,9 @@ func (r GalaxyAdapter) getNextImages(ctx context.Context,
 	// Keep getting the images
 	if iResp.Next != "" {
 		log.Debugf("getting next page of results - %v", iResp.Next)
+		searchURL := fmt.Sprintf(galaxyAPIURL, r.Config.URL.String(), iResp.Next)
 		// Fan out calls to get the next images.
-		go r.getNextImages(ctx, fmt.Sprintf("%v%v", galaxyAPIURL, iResp.Next), ch, cancelFunc)
+		go r.getNextImages(ctx, fmt.Sprintf("%v%v", searchURL, iResp.Next), ch, cancelFunc)
 	}
 	for _, imageName := range iResp.Results {
 		log.Debugf("Trying to load %v.%v", imageName.Summary.Namespace.Name, imageName.Name)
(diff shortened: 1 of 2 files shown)


registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,252 @@
...
+	"net/http"
+	"strings"
+)
+
+const defaultRunner = "ansibleplaybookbundle/apb-base:latest"
+const galaxyName = "galaxy"
+const galaxySearchURL = "https://galaxy-qa.ansible.com/api/v1/content/?content_type__name=apb"
+const galaxyNSSearchURL = "https://galaxy-qa.ansible.com/api/v1/content/?content_type__name=apb&namespace__name=%v"
+const galaxyRoleURL = "https://galaxy-qa.ansible.com/api/v1/content/%v/"
+const galaxyAPIURL = "https://galaxy-qa.ansible.com/api/v1"
```
jmrodri: I would've just grouped these:

```golang
const (
    defaultRunner = "ansibleplaybookbundle/apb-base:latest"
    galaxyName = "galaxy"
    galaxySearchURL = "https://galaxy-qa.ansible.com/api/v1/content/?content_type__name=apb"
    galaxyNSSearchURL = "https://galaxy-qa.ansible.com/api/v1/content/?content_type__name=apb&namespace__name=%v"
    galaxyRoleURL = "https://galaxy-qa.ansible.com/api/v1/content/%v/"
    galaxyAPIURL = "https://galaxy-qa.ansible.com/api/v1"
)
```

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+	}
+	return specs, nil
+}
+
+// getNextImages - will follow the next URL using go routines.
+func (r GalaxyAdapter) getNextImages(ctx context.Context,
+	url string, ch chan<- string,
+	cancelFunc context.CancelFunc) (*GalaxySearchResponse, error) {
+	req, err := http.NewRequest("GET", url, nil)
```
jmrodri: can we add a blank line after the { and before the `req, err`?

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	spec = roleResp.Metadata.Spec
+	spec.Runtime = 2
+
+	if len(r.Config.Runner) == 0 {
+		spec.Image = defaultRunner
+	} else {
+		spec.Image = r.Config.Runner
+	}
```
jmrodri: I would change this to be:

```golang
spec.Image = defaultRunner
if len(r.Config.Runner) != 0 {
    spec.Image = r.Config.Runner
}
```

jmrodri (changes requested):
a few minor changes.

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+	//yaml "gopkg.in/yaml.v1"
+	"io/ioutil"
+	"net/http"
+	"strings"
+)
+
+const (
+	defaultRunner     = "ansibleplaybookbundle/apb-base:latest"
+	galaxyName        = "galaxy"
+	galaxySearchURL   = "https://galaxy-qa.ansible.com/api/v1/content/?content_type__name=apb"
```
eriknelson: I'm guessing there is a reason for this that I'm not privy to, but why are these hardcoded? Shouldn't the adapter be written with this parameterized?

eriknelson: +1 to making these variables though.

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+
+// RegistryName - Retrieve the registry name
+func (r GalaxyAdapter) RegistryName() string {
+	return galaxyName
+}
+
+// GetImageNames - retrieve the images
+func (r GalaxyAdapter) GetImageNames() ([]string, error) {
+	log.Debug("GalaxyAdapter::GetImages")
+	log.Debug("BundleSpecLabel: %s", BundleSpecLabel)
```
jmrodri: should be `log.Debugf` since you are using a formatting string.

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+
+package adapters
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/automationbroker/bundle-lib/bundle"
+	log "github.com/sirupsen/logrus"
+	//yaml "gopkg.in/yaml.v1"
```
jmrodri: remove this yaml comment.

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+	log.Debug("BundleSpecLabel: %s", BundleSpecLabel)
+	log.Debug("Loading role list with tag: [apb]")
+
+	channel := make(chan string)
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+
+	var imageResp *GalaxySearchResponse
+	var err error
+	// Intial call to getNextImages this will fan out to retrieve all the values.
```
jmrodri: typo: Intial should be "Initial". Also a comma after getNextImages.

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+		imageResp, err = r.getNextImages(ctx, galaxySearchURL, channel, cancelFunc)
+	} else {
+		imageResp, err = r.getNextImages(ctx, fmt.Sprintf(galaxyNSSearchURL, r.Config.Org), channel, cancelFunc)
+	}
+	// if there was an issue with the first call, return the error
+	if err != nil {
+		return nil, err
+	}
+
+	// If no results in the fist call then close the channel as nothing will get loaded.
```
jmrodri: typo: fist should be first

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+		close(channel)
+	}
+	var apbData []string
+	counter := 1
+	for imageData := range channel {
+		apbData = append(apbData, imageData)
+		if counter < imageResp.Count {
+			counter++
+		} else {
+			close(channel)
```
jmrodri, line 133: This feels weird to do in a for loop. But I guess it's the only way to exit the loop. Just spidey senses, nothing concrete wrong with it.

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+// getNextImages - will follow the next URL using go routines.
+func (r GalaxyAdapter) getNextImages(ctx context.Context,
+	url string, ch chan<- string,
+	cancelFunc context.CancelFunc) (*GalaxySearchResponse, error) {
+
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		log.Errorf("unable to get next roles for url: %v - %v", url, err)
+		cancelFunc()
+		close(ch)
```
jmrodri, line 176: are you supposed to close the channel here? seems like the person calling you should do the closing. That is, you did not create the resource so you shouldn't be destroying it.

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+		return nil, err
+	}
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		log.Errorf("unable to get next roles for url: %v - %v", url, err)
+		cancelFunc()
+		close(ch)
```
jmrodri, line 184: same comment as above about closing channel you didn't create.

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+	defer resp.Body.Close()
+
+	imageList, err := ioutil.ReadAll(resp.Body)
+
+	iResp := GalaxySearchResponse{}
+	err = json.Unmarshal(imageList, &iResp)
+	if err != nil {
+		log.Errorf("unable to get next images for url: %v - %v", url, err)
+		cancelFunc()
+		close(ch)
```
jmrodri, line 196: same comment as above regarding closing of channel.

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+		return &iResp, err
+	}
+	// Keep getting the images
+	if iResp.Next != "" {
+		log.Debugf("getting next page of results - %v", iResp.Next)
+		// Fan out calls to get the next images.
+		go r.getNextImages(ctx, fmt.Sprintf("%v%v", galaxyAPIURL, iResp.Next), ch, cancelFunc)
```
jmrodri: where does the response go?

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+			}
+		}(imageName)
+	}
+	return &iResp, nil
+}
+
+func (r GalaxyAdapter) loadSpec(imageName string) (*bundle.Spec, error) {
+	imageSplit := strings.Split(imageName, "#")
+	roleName := imageSplit[0]
+	roleID := imageSplit[1]
```
jmrodri: if `imageName` does not have a # in it, we will panic here. We should really check the length of `imageSplit`

jmrodri (changes requested):
some typos, some concerns over closing channels, and the Split will panic if there is no # in the imageName.

registries/adapters/galaxy_adapter.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	// If no results in the fist call then close the channel as nothing will get loaded.
+	if len(imageResp.Results) == 0 {
+		log.Info("canceled retrieval as no items in org")
+		close(channel)
+	}
+	var apbData []string
+	counter := 1
+	for imageData := range channel {
```
eriknelson, line 128: I'm trying to reason through the code to determine if it's possible that there have not been enough responses back to this channel to tick the counter high enough to reach the expected count, while also not having had the channel closed by the inner function. In that case, I think this would dead lock as written.

registries/adapters/galaxy_adapter.go
eriknelson: I'm a little concerned as well? I think of the channel as a shared resource that should have one owner. It's not uncommon to have an outer fn make a channel and then pass it to an inner fn (effectively passing ownership over to the inner fn), and then for looping over the channel range until the inner fn has closed the channel. It feels a little weird to me here that both the outer, and inner fn seem to claim ownership over when this channel should be closed, thus ending the loop.

registries/adapters/galaxy_adapter.go
```diff
@@ -180,8 +192,9 @@ func (r GalaxyAdapter) getNextImages(ctx context.Context,
 	// Keep getting the images
 	if iResp.Next != "" {
 		log.Debugf("getting next page of results - %v", iResp.Next)
+		searchURL := fmt.Sprintf(galaxyAPIURL, r.Config.URL.String(), iResp.Next)
 		// Fan out calls to get the next images.
-		go r.getNextImages(ctx, fmt.Sprintf("%v%v", galaxyAPIURL, iResp.Next), ch, cancelFunc)
+		go r.getNextImages(ctx, fmt.Sprintf("%v%v", searchURL, iResp.Next), ch, cancelFunc)
```
eriknelson: Looking at the response body for a paged request with a next (https://galaxy-qa.ansible.com/api/v1/content/?content_type__name=apb&page_size=2), this is going to request a bad URL:

`"next": "/content/?content_type__name=apb&page=2&page_size=2",`

The `searchURL` here already has the path in it, so you're going to end up with something like:

`https://galaxy-qa.ansible.com/api/v1/content/?content_type__name=apb/content/?content_type__name=apb&page=2&page_size=2`

I'm thinking these fan out calls should do something like `fmt.Sprintf("%v%v", r.Config.URL, iResp.Next)`, since the next value contains the full next path, so we just need to tack the host onto the front of it.

eriknelson (changes requested):

registries/adapters/galaxy_adapter.go
djzager:
```
searchURL := fmt.Sprintf(galaxyAPIURL, r.Config.URL.String(), iResp.Next)
```

My belief is that I'm constructing the `searchURL` correctly, 
That should take the host `https://galaxy-qa.ansible.com` + `/api/v1` + `/content/?content_type__name=apb&page=2&page_size=2`. That looks correct to me.

I think my mistake was simply not using the `searchURL`:

```
go r.getNextImages(ctx, searchURL, ch, cancelFunc)
```

shawn-hurley (approved):
I believe this will behave as expected. As mentioned offline by @eriknelson if we really wanted to speed up the process we would have to make assumptions about how to get pages, determine the number of pages and asynchronously get each page.

jmrodri:
We can ignore the coveralls decrease in test coverage for now. I am writing adapter tests and I'll include one for galaxy when it gets committed.

jmrodri (approved):

eriknelson (approved):

PR #131, opened 2018-07-09
random deprovision error about DestroySandbox

jmrodri:
looking through the test this was the only test that had a mock for
CreateSandbox but did not have an expectation for DestroySandbox.


jmrodri:
It's still happening, this wasn't it.

PR #132, opened 2018-07-09
add API V2 schema 2 support for apiv2 adapter

johnkim76:


johnkim76:
need to add a test for schema2 and test it

registries/adapters/adapter.go
```diff
@@ -179,3 +182,14 @@ func getAPBRuntimeVersion(version string) (int, error) {
...
+// manifest schema version
+func getSchemaVersion(response []byte) (int, error) {
+	mResp := manifestResponse{}
+	r := bytes.NewReader(response)
+	if err := json.NewDecoder(r).Decode(&mResp); err != nil {
+		log.Errorf("Error getting schemaVersion", err)
+		return 0, err
+	}
+	return mResp.SchemaVersion, nil
+}
```
jmrodri, line 195: this feels weird we're parsing the response bytes multiple times from bytes to json. We do it in `getSchemaVersion`. And if it version 1 we do it again in `responseToSpec`. reading the same set of bytes parsing the json and loading the same response struct. Feels like once we've loaded it we should stick with it.

registries/adapters/apiv2_adapter.go
```diff
@@ -262,5 +263,43 @@ func (r APIV2Adapter) loadSpec(imageName string) (*bundle.Spec, error) {
...
+		}
+		resp, err = r.client.Do(req)
+		if err != nil {
+			return nil, err
+		}
+		body, err = registryResponseHandler(resp)
+		if err != nil {
+			return nil, fmt.Errorf("%s - error getting configuration object for image [%s] : %s", r.config.AdapterName, imageName, err)
+		}
+		return configToSpec(body, fmt.Sprintf("%s/%s:%s", registryName, imageName, r.config.Tag))
```
jmrodri, line 301: Is there a reason the parsing for v2 is done inline above, while v1 is all done in a method in adapter.go?

jmrodri (approved):
I have some questions, but nothing that will hold this up.

registries/adapters/adapter.go
```diff
@@ -86,87 +106,70 @@ func registryResponseHandler(resp *http.Response) ([]byte, error) {
...
-		Spec             string `json:"com.redhat.apb.spec"`
-		LegacyAPBRuntime string `json:"com.redhat.apb.runtime"`
-		BundleRuntime    string `json:"com.redhat.bundle.runtime"`
-	}
-
-	type config struct {
-		Label label `json:"Labels"`
-	}
+// Retrieve the spec from a manifest response
+func responseToSpec(response []byte, image string) (*bundle.Spec, error) {
```
dymurray, line 110: +1

registries/adapters/apiv2_adapter.go
johnkim76: For schema v1, we can just read the body, but in schema v2, we need to digest, which is specific for each image, from the 1st response to get the repo config.

dymurray: I think intention was that the other adapters could continue to use `responseToSpec` as they were in the past while schema `2` requires grabbing the `config` section from a different endpoint.

dymurray (approved):
LGTM. Nice job on the tests.

registries/adapters/apiv2_adapter.go
```diff
@@ -262,5 +263,43 @@ func (r APIV2Adapter) loadSpec(imageName string) (*bundle.Spec, error) {
 		registryName = fmt.Sprintf("%s:%s", r.config.URL.Hostname(), r.config.URL.Port())
 	}
 
-	return imageToSpec(body, fmt.Sprintf("%s/%s:%s", registryName, imageName, r.config.Tag))
+	schemaVersion, err := getSchemaVersion(body)
+	if err != nil {
+		return nil, err
+	}
+
+	switch schemaVersion {
```
shawn-hurley, line 271: Why can't this entire block move into responseToSpec? that way everyone can get this behavior

shawn-hurley (changes requested):

registries/adapters/apiv2_adapter.go
johnkim76: @dymurray actually asked me the same thing...and I started out implementing it that way. However, I didn't see a good way to have the authenticated client available in responseToSpec, and I this functionality is specific only to API V2 compliant registries.

shawn-hurley (approved):

Issue #120, opened 2018-07-05
bundle.Spec needs a mark for deletion field


alaypatel07:
With the approach for reconciliation decided [here](https://www.redhat.com/archives/ansible-service-broker/2018-July/msg00003.html), the bundle.Spec struct needs to have a **mark for deletion** field.

jmrodri:
@shawn-hurley mentioned kubernetes objects already have this field in the ObjectMeta
https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta

Issue #10, opened 2018-03-15
Move WatchPod into runtime


djzager:
Checklist:

- [x] Move WatchPod into the runtime package #1 
- [x] Make WatchPod more agnostic #3

PR #137, opened 2018-07-11
Deletion field

alaypatel07:
Add a `Delete bool` field to Spec struct in bundle/types.go. This PR addresses issue [#120](https://github.com/automationbroker/bundle-lib/issues/120).  

The PR depends on pull [#10](https://github.com/automationbroker/broker-client-go/pull/10). Gopkg.toml and Gopkg.lock will change when it gets merged

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -22,8 +22,7 @@
 
 [[constraint]]
   name = "github.com/automationbroker/broker-client-go"
-  branch = "deletion-field"
-  source = "github.com/alaypatel07/broker-client-go"
+  branch = "master"
 
 [[constraint]]
   name = "github.com/coreos/etcd"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -21,8 +21,9 @@
 
 
 [[constraint]]
-  version = "0.1.2"
   name = "github.com/automationbroker/broker-client-go"
+  branch = "deletion-field"
+  source = "github.com/alaypatel07/broker-client-go"
 
 [[constraint]]
   name = "github.com/coreos/etcd"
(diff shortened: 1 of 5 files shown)


Gopkg.toml
```diff
@@ -21,8 +21,9 @@
 
 
 [[constraint]]
-  version = "0.1.2"
   name = "github.com/automationbroker/broker-client-go"
+  branch = "deletion-field"
+  source = "github.com/alaypatel07/broker-client-go"
```
eriknelson: We'll have to make sure this gets updated once the broker-client-go PR merges.

bundle/types.go
```diff
@@ -152,6 +152,19 @@ type Spec struct {
 	Async       string                 `json:"async"`
 	Plans       []Plan                 `json:"plans"`
 	Alpha       map[string]interface{} `json:"alpha,omitempty"`
+	Delete      bool                   `json:"delete"`
+}
+
+// MarkDelete - marks the spec for deletion, which indicates that the spec
+// will eventually be deleted from the datastore
+func (s *Spec) MarkDelete() {
```
shawn-hurley: I personally do not prefer these helper methods. Delete being a bool I think the Caller could easily update it IMO.

shawn-hurley (changes requested):

bundle/types.go
eriknelson: +1. If this were likely to change in the future it would warrant a fn, but I don't think this is worth it.

bundle/types.go
```diff
@@ -152,6 +152,19 @@ type Spec struct {
...
+	s.Delete = true
+}
+
+// UnMarkDelete - un-marks the spec for deletion, which indicates that the
+// spec is in the registry and will be saved in datastore
+func (s *Spec) UnMarkDelete() {
```
eriknelson: Also not necessary.

eriknelson (changes requested):

jmrodri (approved):
LGTM

eriknelson (approved):

shawn-hurley (approved):

PR #142, opened 2018-07-18
Only create network policy when using transient namespace

dymurray:
This PR is needed for `apb` tool because some clusters do not allow the end user to create Network Policies. There really is no need to create one if the user is deploying the APB directly into the target namespace. I simply moved the creation logic inside of the conditional that checks if the target NS is the same as the transient NS.

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -262,8 +262,8 @@ func (p provider) CreateSandbox(podName string,
 		// Create namespace.
 		ns := &apicorev1.Namespace{
 			ObjectMeta: metav1.ObjectMeta{
-				Labels:       metadata,
-				GenerateName: namespace,
+				Labels: metadata,
+				Name:   namespace,
 			},
 		}
 		ns, err = k8scli.Client.CoreV1().Namespaces().Create(ns)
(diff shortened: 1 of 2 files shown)


djwhatle:
Reason for this change as I understand it:
 - Creating Network Policies requires a fairly high tier of permissions in OpenShift
 - We sometimes want to operate inside of a single namespace (e.g. with `apb provision`) in a brokerless env
 - We should skip the step of creating the Network Policy when possible, because it allows us to lower the overall permissions required to work with APBs.

Edit:
Just updated the page and saw that Dylan updated the original comment, making this one redundant. Oh well :)

djwhatle (approved):
LGTM, and did a quick test with apb tool.

shawn-hurley:
@dymurray Do you think you have time to add a test case for the sandbox creation? Also you may want to add the same check for the deletion of sandbox. No need to add error logs if we expect it to error?

dymurray:
@shawn-hurley In order to get the tests to stop failing (and I don't know why) I had to change `GenerateName` to `Name`. For some reason the fake client was trying to create a namespace with an empty name if `Name` wasn't set.

Do you see any problems doing this in runtime.go?

dymurray:
Coverage now up 2.4%

shawn-hurley:
I would prefer to not have to do this. I like the extra that kube gives us
and allows us to guarantee uniqueness which I think is important.

I’ll go see if there is a way to make the fake client work better for us in
this use case.

dymurray:
Added a reactor to handle generateName coming from `runtime.go`

shawn-hurley (approved):
LGTM, thanks for the test case!

PR #144, opened 2018-07-20
Add dockerhub_adapter tests to increase coverage

jmrodri:


jmrodri:
`dockerhub_adapter.go` has 0.8% coverage in `master`. The above tests increase that coverage to 76.0%. Covers most of the happy path.

registries/adapters/dockerhub_adapter.go
```diff
@@ -222,24 +220,30 @@ func (r DockerHubAdapter) loadSpec(imageName string) (*bundle.Spec, error) {
...
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		return nil, err
 	}
+
 	body, err := registryResponseHandler(resp)
 	if err != nil {
 		return nil, fmt.Errorf("DockerHubAdapter::error handling dockerhub registery response %s", err)
 	}
+	fmt.Printf("body: [%v], %s\n", string(body), fmt.Sprintf("%s/%s:%s", r.RegistryName(), imageName, r.Config.Tag))
```
shawn-hurley: Should we log this at debug level? or should we just remove it?

registries/adapters/dockerhub_adapter.go
```diff
@@ -222,24 +220,30 @@ func (r DockerHubAdapter) loadSpec(imageName string) (*bundle.Spec, error) {
 	if r.Config.Tag == "" {
 		r.Config.Tag = "latest"
 	}
-	req, err := http.NewRequest("GET", fmt.Sprintf(dockerHubManifestURL, imageName, r.Config.Tag), nil)
+
+	token, err := r.getBearerToken(imageName)
```
shawn-hurley, line 224: It looks like the order of these calls has changed, is there a reason?

shawn-hurley (changes requested):

registries/adapters/dockerhub_adapter.go
dymurray: Seems to make sense to me. Now all the request logic is grouped together. Better to grab the token and error early before constructing the request.

registries/adapters/dockerhub_adapter.go
dymurray: +1 I think this will play ugly with the `apb` tool.

dymurray (changes requested):
Great job on the tests, learned something new :)

Only thing I think should change is the print statement to be a log statement.

registries/adapters/dockerhub_adapter.go
jmrodri: oops this was not supposed to be there. I usually put in XXX and forgot to. I will be removing this line.

dymurray (approved):

jmrodri:
@dymurray @shawn-hurley removed the rogue printf statement. thanks for catching that.

registries/adapters/dockerhub_adapter.go
jmrodri: @shawn-hurley yes, I changed the order on purpose. Before we would create the request, then go out and get the token, THEN use the request we created earlier. It seemed confusing to me. Felt better to get the token. If that failed, no need to create the request. 

Also, I wanted the request creation close to where it was called to illustrate that it has no bearing on the bearer token call.

PR #150, opened 2018-07-24
only add mode: count if it doesn't already exist

jmrodri:


jmrodri:
Fixes test-coverage-html target, we were sometimes putting multiple mode: count headers which caused it to error out

djzager (approved):

PR #156, opened 2018-07-27
Add galaxy_url parameter to object

djzager:
Since the galaxy URL is configurable, we need to pass the server URL into the APB at runtime.


eriknelson (approved):
LGTM

djzager:
I was able to build a broker and test this with `apb-base:canary`. The broker properly passes the `galaxy_url` and the updated `apb-base` uses that URL for downloading the role.

jmrodri (approved):
VISIACK

PR #160, opened 2018-08-03
Bug 1609700 - Use config user/pass if no auth_type

eriknelson:
A regression was introduced with the quay adapter that removed these and broke default logins with user+pass and no `auth_type`. This adds them back in.


jmrodri (changes requested):
Please add a test for this. Probably can change this one to be more correct:
https://github.com/automationbroker/bundle-lib/blob/master/registries/registry_test.go#L1062-L1074

eriknelson:
The bug that this PR fixes was effectively stripping the User/Pass combo. The updated test ensures this doesn't happen.

djzager (approved):

jmrodri (approved):

PR #161, opened 2018-08-08
Bug 1583049 - Add metrics

eriknelson:
* Updates Gopkg.lock
* Fixes tests based on new deps

### DIFF
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -1,0 +1,91 @@
+//
+// Copyright (c) 2018 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


djzager (approved):

metrics/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+		err := prom.Register(collector)
+		if err != nil {
+			log.Errorf("unable to register collector with prometheus: %v", err)
+		}
+	})
+}
+
+// SandboxCreated - Counter for how many sandbox created.
+func SandboxCreated() {
+	if collector != nil {
```
shawn-hurley: I think that I would want to know if the collector is nil and I am not collecting metrics. I think the `recover..Panic` should capture that so we may want to remove this?

shawn-hurley (changes requested):
Small nit

shawn-hurley (approved):

PR #164, opened 2018-08-15
Bug 1613238 - Fix oauth client proxy

eriknelson:
Since our oauth client is configuring its own Transport, it's overriding the default proxy strategy. This fix ensures we explicitly consume the proxy config from the environment, which is normally the default Transport's proxy strategy.


jmrodri (approved):
ACK

mhrivnak (approved):

PR #167, opened 2018-08-28
Bug 1613238 - Ensure we pass a path to client.NewRequest

eriknelson:
We had an inconsistency where NewRequest was used in the apiv2 adapter,
we were passing in a full URL including the hostname instead of just the
path (which is what NewRequest expected). This caused the scheme://host
to be prefixed twice. Apparently http.client.Do was able to sanitize
this in some way with an implicit Parse call, but the problem came to
light when the URL was passed to a proxy to execute, and it wasn't
cleaned.

### DIFF
--- a/registries/adapters/apiv2_adapter.go
+++ b/registries/adapters/apiv2_adapter.go
@@ -29,10 +29,10 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


jmrodri (approved):
seems to make sense

dymurray (approved):
Worked for me outside of proxy env

PR #170, opened 2018-08-31
Bug 1623331 - Consider empty proxy vars as not configured.

jmrodri:
Setting HTTP(S)_PROXY to empty string will be considered as present in
the environment even if they are empty. But having empty values means
there is no proxy. This could be considered a misconfiguration so we log
a warning. Then we return nil as we don't want to set an improper proxy
configuration.

### DIFF
--- a/bundle/executor_test.go
+++ b/bundle/executor_test.go
@@ -18,7 +18,11 @@ package bundle
 
 import (
 	"errors"
+	"os"
 	"testing"
+
+	"github.com/automationbroker/bundle-lib/runtime"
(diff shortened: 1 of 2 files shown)


bundle/executor_test.go
```diff
@@ -91,3 +95,88 @@ func TestExecutor(t *testing.T) {
 		})
 	}
 }
+
+func TestGetProxyConfig(t *testing.T) {
```
eriknelson, line 99: :tada:

eriknelson (approved):

dymurray (approved):

PR #171, opened 2018-09-06
Bug 1625969 - Add metadata label check for local registry

dymurray:
This was a weird bug. Basically the RH build system is putting the APB spec in the images metadata that populates itself in the OpenShift API differently. For some reason when you retag the images from the puddle and put them in the internal registry, the `containerConfig` section of the metadata has the `com.redhat.apb.runtime` label not but the `com.redhat.apb.spec` label. The `config` section of the metadata has both labels.

This is different than if you were to follow our recommended development guide where you create buildconfigs and imagestreams yourself. Because of this I put a check to see if the label exists in `containerConfig` and if not check `Config`. 

The other nice thing about this fix is that before we weren't even checking if the label was set to `""` so this will error out nicely even if the spec is missing in both.

### DIFF
--- a/registries/adapters/local_openshift_adapter.go
+++ b/registries/adapters/local_openshift_adapter.go
@@ -42,6 +42,7 @@ type containerConfig struct {
 
 type imageMetadata struct {
 	ContainerConfig containerConfig `json:"ContainerConfig"`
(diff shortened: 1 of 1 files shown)


jmontleon (approved):
ACK.

PR #172, opened 2018-10-05
Bug 1613280 - only create netpolicy if one already exists

jmrodri:
If there are other services in the target namespace talking to each other with no network policy, during APB deployments we automatically create one that blocks the other services. This change looks to see if there are any network policies on the target namespace. If there are none, we forgo creating the network policy, assuming things are open and the transient namespace should be able to talk to the target with no issues. If there are existing network policies, then we will add ours to give the transient namespace access to the target. There is still the chance that our network policy could still affect things depending on the variety of existing network policies in place, too many to verify.

### DIFF
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/automationbroker/bundle-lib/clients"
 	"github.com/automationbroker/bundle-lib/runtime/mocks"
 	apicorev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/fake"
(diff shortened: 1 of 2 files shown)


shawn-hurley (approved):
This looks good to me I don't see any problems

VISACK

If you want a second pair of eyes on testing let me know and I can spin up my dev env for this again

runtime/runtime_test.go
```diff
@@ -88,23 +89,52 @@ func TestCreateSandbox(t *testing.T) {
...
-			name:      "Test Create Sandbox with namespace not in target",
+			name:      "Test Create Sandbox with ns not in target no net policy",
 			podName:   "pod-name",
 			client:    fake.NewSimpleClientset(),
 			namespace: "bar-ns",
 			targets:   []string{"satoshi-ns", "nakamoto-ns"},
 			apbRole:   "edit",
 		},
+		{
+			name:      "Test Create Sandbox with ns in target with net policy",
```
eriknelson, line 111: :tada:

eriknelson (approved):

PR #175, opened 2019-01-21
Added support for additional token field during Docker registry authorization

### LINKED ISSUE #174: Respect additional authorization token field
Docker specifies the following `token` headers during registry authorization:
- `token`: An opaque Bearer token that clients should supply to subsequent requests in the Authorization header.
- `access_token`: For compatibility with OAuth 2.0, we will also accept token under the name access_token. At least one of these fields must be specified, but both may also appear (for compatibility with older clients). When both are specified, they should be equivalent; if they differ the client's choice is undefined.

See https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md#token-response-fields for more details.

The `bundle-lib` only accepts the `access_token` field. See https://github.com/automationbroker/bundle-lib/blob/master/registries/adapters/oauth/client.go#L35-L37 and https://github.com/automationbroker/bundle-lib/blob/master/registries/adapters/oauth/client.go#L191 for more details. 

But registries like Artifactory or Harbor send their token in the field `token`. Therewith we couldn't connect them with the Automation Broker.

I'll create a  PR which respects the additional token field `token` as well.

sjahreis:
See #174 for more details.

Closes #174.

### DIFF
--- a/registries/adapters/oauth/client_test.go
+++ b/registries/adapters/oauth/client_test.go
@@ -35,6 +35,20 @@ var headerErrorCases = map[string]string{
 	"Bearer realm=\"\"":      "",
 }
 
+var tokenCases = map[string]string{
+	"{\"access_token\": \"abc123\"}":                        "abc123",
+	"{\"token\": \"abc123\"}":                               "abc123",
+	"{\"access_token\": \"abc123\", \"token\": \"def456\"}": "abc123",
+	"{}": "",
+}
+
+var tokenErrorCases = map[string]string{
+	"{\"token\": {}":          "unexpected end of JSON input",
+	"{\"access_token\": {}":   "unexpected end of JSON input",
+	"{\"token\": null":        "unexpected end of JSON input",
+	"{\"access_token\": null": "unexpected end of JSON input",
+}
+
 func TestParseAuthHeader(t *testing.T) {
 	for in, out := range headerCases {
 		result, err := parseAuthHeader(in)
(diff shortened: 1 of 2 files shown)


eriknelson:
Patch looks good, thanks for the contribution. Going to run some tests with this but I think we can get this merged today and into ASB master. Where are you sourcing your broker images from? You may need to build yourself an image and override your deployment to use that until we're able to release.

sjahreis:
@eriknelson: Thanks! 😁

We are using the official ASB images from Dockerhub. We might switch to a own image. Can you estimate when the next release will be published?

registries/adapters/oauth/client_test.go
```diff
@@ -58,6 +72,29 @@ func TestParseAuthHeaderErrors(t *testing.T) {
 	}
 }
 
+func TestParseAuthToken(t *testing.T) {
```
shawn-hurley, line 75: 👍 🎉

shawn-hurley (approved):
I wonder if we could inline the logic.

Create a single struct like:

```go
type oauth2Response struct {
		AccessToken string `json:"access_token"`
		Token       string `json:"token"`
}
```

Then you can have a single Unmarshal and look at which is defined?

@jmrodri you have thoughts


This is documented and clean so that this is not a make or break it for me, therefore it LGTM

eriknelson:
Going to merge this so we can try and get an image released for you today.

eriknelson (approved):

PR #182, opened 2019-06-29
Bug 1723422 - properly handle pagination url

jmrodri:
* set path and query string properly
* fix getNextImageURL parsing of path url
* add unit test for getNextImageURL

### DIFF
--- a/registries/adapters/oauth/client.go
+++ b/registries/adapters/oauth/client.go
@@ -81,7 +81,16 @@ func (c *Client) NewRequest(path string) (*http.Request, error) {
 	if err != nil {
 		return nil, err
 	}
-	req.URL.Path = path
+
+	// path might be a fully qualified URL and we only want the path
+	// and query bits
+	thebits, err := url.Parse(path)
+	if err != nil {
+		return nil, err
+	}
+
+	req.URL.Path = thebits.Path
+	req.URL.RawQuery = thebits.Query().Encode()
 	if c.token != "" {
 		req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", c.token))
 	}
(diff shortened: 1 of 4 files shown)


registries/adapters/oauth/client.go
```diff
@@ -81,7 +81,16 @@ func (c *Client) NewRequest(path string) (*http.Request, error) {
 	if err != nil {
 		return nil, err
 	}
-	req.URL.Path = path
+
+	// path might be a fully qualified URL and we only want the path
+	// and query bits
+	thebits, err := url.Parse(path)
```
shawn-hurley: personally, I would change the variable name but not a blocker

registries/adapters/oauth/client.go
```diff
@@ -81,7 +81,16 @@ func (c *Client) NewRequest(path string) (*http.Request, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	req.URL.Path = thebits.Path
```
shawn-hurley: could we just set `req.URL = thebits`?

shawn-hurley (approved):
I think it will fix the problem, just some nits.

registries/adapters/oauth/client.go
jmrodri: The comment on `NewRequest` states they use the base URL from the client. So I wanted to keep the same functionality. 
```golang
// NewRequest - creates and returns a *http.Request assuming the GET method.
// The base URL configured on the Client gets used with its Path component
// replaced by the path argument. If a token is available, it is added to the
// request automatically. "Accept: application/json" is added to all requests.
// The caller should customize the request as necessary before using it.
```

registries/adapters/oauth/client.go
jmrodri: changed to `pathAsUrl` let me know if that is not okay.

jmrodri: actually changed it to `pathAsURL` since `pathAsUrl` pissed off lint :)

PR #184, opened 2019-10-18
Bug 1763005 - add registry-proxy support

jmrodri:
The registry-proxy GET /v2 returns a WWW-Authenticate header with only a
realm: .../v2/auth the APIV2Adapter simply gets a token for /v2/auth
which is not valid for registry-proxy which seems to require a scoped
token.

In this PR, we added a new RegistryProxyAdapter that now requires the
apbs to be listed in the configuration. For each of those images listed,
we will add them to the scope when we ask for the token. This allows the
broker access to fetch the apb images.

Also, added a test to ensure registry proxy errors when no images are defined
and added a unit test for getTokenWithScope.

### DIFF
--- a/registries/adapters/apiv2_adapter.go
+++ b/registries/adapters/apiv2_adapter.go
@@ -80,6 +80,7 @@ func NewPartnerRhccAdapter(config Configuration) (PartnerRhccAdapter, error) {
 	return PartnerRhccAdapter{apiV2}, nil
 }
 
+// NewRegistryProxyAdapter - create a new Registry Proxy Adapter
 func NewRegistryProxyAdapter(config Configuration) (RegistryProxyAdapter, error) {
 	// we want to use a different token for this adapter, so do not call
 	// NewAPIV2Adapter directly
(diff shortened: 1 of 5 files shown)


registries/adapters/apiv2_adapter.go
```diff
@@ -75,6 +80,35 @@ func NewPartnerRhccAdapter(config Configuration) (PartnerRhccAdapter, error) {
...
 
+func NewRegistryProxyAdapter(config Configuration) (RegistryProxyAdapter, error) {
+	// we want to use a different token for this adapter, so do not call
+	// NewAPIV2Adapter directly
+	apiv2a := APIV2Adapter{
+		config: config,
+		client: oauth.NewClient(config.User, config.Pass, config.SkipVerifyTLS, config.URL),
+	}
+
+	log.Debugf("XXX number of images configured: %d", len(config.Images))
```
shawn-hurley: remove this, or remove the XXX at least?

shawn-hurley (commented):
looks pretty good to me

@joelanford want to take a look as well?

registries/adapters/apiv2_adapter.go
jmrodri: removed and forced push to keep this commit clean :)

